PR #3, opened 2020-04-30
Configure theme for consistent local preview

ncjones:
I've changed to using the recommended gh pages gem instead of jekyll direcetly which calls the github api to get the repo config that was previously missing.

I've also disabled the links to our github repo on the site by overriding the github configs that are used by the default template and I've added a placeholder logo image.

Note you might need to remove the jekyll metadata cache if this isn't working initially and you might get warnings about github api rate limiting.

See the following for how to configure the theme:

https://github.com/pages-themes/minimal#usage
https://github.com/pages-themes/minimal/blob/master/_layouts/default.html#L17


.gitignore
```diff
@@ -2,3 +2,6 @@ vendor
 _site
 .sass-cache
 .jekyll-metadata
+
+# vim
+*.sw[a-p]
```
maticko, line 7: shouldn't you add this to your `.gitignore_global`?

ncjones: True

PR #11, opened 2020-05-12
Fixed spelling and back to home button

AliceKSanders:
Jerome suggested we should add a button back to the home page (Centrapay website), if someone could make this look nicer? that would be good - since it looks a bit odd at the moment.

### DIFF
--- a/index.md
+++ b/index.md
@@ -6,10 +6,10 @@ title: Centrapay Documentation
 # Introduction
 {:.no_toc}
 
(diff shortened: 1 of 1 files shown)


index.md
```diff
@@ -2,14 +2,15 @@
 layout: default
 title: Centrapay Documentation
 ---
+[Back to home](https://www.centrapay.com){:target="_self"}
```
ncjones, line 5: @AliceKSanders can you move this to a separate PR?

PR #21, opened 2020-06-16
update account creation request and responses

maticko:


accounts.md
```diff
@@ -23,21 +23,28 @@ POST `https://service.centrapay.com/api/accounts`
...
 {
   "id": "Jaim1Cu1Q55uooxSens6yk",
-  "name": "Centrapay Cafe"
+  "name": "Centrapay Cafe",
+  "type": "org",
+  "createdBy": "crn:WIj211vFs9cNACwBb04vQw:api-key:MyApiKey",
+  "createdAt": "2020-06-12T01:17:46.499Z",
+  "modifiedAt": "2020-06-12T01:17:46.499Z",
+  "modifiedBy": "crn:WIj211vFs9cNACwBb04vQw:api-key:MyApiKey",
+  "version": 1
```
ncjones, line 47: I am thinking we should make all integers strings when serialized to JSON (ie versions and currency values). Internally we should switch to using BigInt and implement BigInt.toJSON(): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#Use_within_JSON

ncjones, line 47: I think we should use string for all integer values.

PR #28, opened 2020-07-28
Add Bank authority documentation

daallen303:


authorities.md
```diff
@@ -0,0 +1,126 @@
+# Bank Authorities API
+
+A bank authority represents an individual's consent for Centrapay to transfer funds to and from a bank account. In order to use a Centrapay wallet an individual must have a verified bank authority for that wallet.
```
ncjones, line 3: In order to move funds between a bank account and a Centrapay wallet, there must be a verified bank authority for the Centrapay account that the wallet belongs to.

PR #51, opened 2020-11-04
Asset Transfer instructions

heymatthew:


src/fiat/asset-transfers.md
```diff
@@ -0,0 +1,30 @@
...
+* TOC
+{:toc}
+
+## Create an Asset Transfer
+
+{% endpoint POST https://service.centraypay.com/api/asset-transfer-requests %}
+
+```sh
+curl -X POST "https://service.centraypay.com/api/asset-transfer-requests" \
+  -H "x-api-key: 1234"
```
heymatthew, line 22: Just wondering if this should mention or prefer `Authorization:` headers?

ncjones: We need to add a separate section to explain the different approaches for authenticating API requests. I think it is more natural to use api keys as the default auth method in the docs. We can publish an sdk that makes it easy to make api calls using auth tokens instead. We also need to document our roles and permissions.

ncjones: ...but all of this can come later

src/fiat/asset-transfers.md
```diff
@@ -0,0 +1,34 @@
...
+  -H "content-type: application/json" \
+  -d '{
+    assetType: "centrapay.nzd.main",
+    assetId: "YGRo6TYYSxH3js7",
```
heymatthew, line 26: If asset type is "centrapay.nzd.main", should the assetId be required and set to walletId?

Given we've got `ledgerId` on /api/wallets, what's a use case which `assetType` wouldn't be ledgerId? e.g. Mad Mike Dollars

If we don't have to future proof, should we use `ledgerId` here?

ncjones: The intention is that asset id is conditionally required depending on the asset type. If the asset type is ledger-based, like "centrapay.nzd.main", then it would not be required. If it is an atomic/discreet asset like a voucher then the asset id would be required.

heymatthew: Fixed in a pairing session 💖💖💖

ncjones (approved):

PR #54, opened 2020-11-08
Remove asset xfer "claimedAt" and "sentAt"

ncjones:
These fields will be persisted with the action log so are not exposed by default. We can expose them in the API later when needed.


heymatthew (approved):
LGTM!

PR #65, opened 2020-11-27
add asset endpoints

maticko:


src/assets/discrete-assets.md
```diff
@@ -77,3 +77,73 @@ curl -X POST "https://service.centrapay.com/api/external-assets" \
...
+  "expiresAt": "2020-12-31T00:00:00.000Z",
+  "createdAt": "2020-05-01T12:30:00.000Z"
+}]
+```
+
+**Error Responses**
+
+| Status |          Code           |                           Description                            |
+|:-------|-:-----------------------|:-----------------------------------------------------------------|
+| 403    | {% break _ FORBIDDEN %} | Invalid account id or account id you don't have permissions for. |
```
ncjones, line 149: This is a standard error code that can be returned from every non-public endpoint so we don't need to document it.

src/assets/discrete-assets.md
```diff
@@ -77,3 +77,73 @@ curl -X POST "https://service.centrapay.com/api/external-assets" \
...
+
+```sh
+curl -X GET "https://service.centrapay.com/api/accounts/Te2uDM7xhDLWGVJU3nzwnh/assets" \
+  -H "x-api-key: 1234"
+```
+
+**Example response payload**
+
+```json
+[{
```
ncjones, line 130: This endpoint will be paginated so the response will be wrapped with a page object:

```
{
  "items": []
}
```

src/assets/discrete-assets.md
```diff
@@ -77,3 +77,73 @@ curl -X POST "https://service.centrapay.com/api/external-assets" \
 | Status | Code                    | Description                                         |
 |:-------|:------------------------|:----------------------------------------------------|
 | 403    | {% break _ FORBIDDEN %} | Invalid external id or pin or asset already exists. |
+
+## Get Asset
```
ncjones, line 81: These two endpoints are not specific to discrete assets or wallets so I am not sure where to put them. What about the parent page?

ncjones (approved):

daallen303 (approved):

PR #68, opened 2020-11-30
Cleanup transacting doc

ncjones:
Hard-wrapped some long lines, added internal links for endpoints and
fixed some typos.


heymatthew (approved):
LGTM, way better with those line breaks

PR #70, opened 2020-12-08
update discrete-asset docs with error messages

Ray2126:


src/assets/discrete-assets.md
```diff
@@ -75,4 +75,8 @@ curl -X POST "https://service.centrapay.com/api/external-assets" \
 
 | Status | Code                    | Description                                         |
 |:-------|:------------------------|:----------------------------------------------------|
-| 403    | {% break _ FORBIDDEN %} | Invalid external id or pin or asset already exists. |
+| 500    | SERVICE_UNAVAILABLE     | Upstream server unavailable.                        |
```
ncjones, line 78: We don't need to document 5xx errors as they can always be handled the same way.

src/assets/discrete-assets.md
```diff
@@ -75,4 +75,8 @@ curl -X POST "https://service.centrapay.com/api/external-assets" \
...
+| 403    | UNKNOWN_CARD            | Card number or PIN is incorrect.                    |
+| 403    | DUPLICATE_ASSET         | External asset already claimed by another account.  |
```
ncjones, line 80: There's also "UNSUPPORTED_CARD_TYPE". I think it should be changed to "UNSUPPORTED_ASSET_TYPE" though.

ncjones: hmm, "UNKNOWN_CARD" would need to be changed to "UNKNOWN_ASSET" for consistency though.

Ray2126: I agree with using asset over card here because the document header is external asset. We also document the response payload with type, which could be more than just a card.

src/assets/discrete-assets.md
Ray2126: Cool, I'll remove this.

src/assets/discrete-assets.md
Ray2126: Will also need to update this in the backend if you are happy with it

ncjones (approved):

PR #77, opened 2021-01-07
Add: centrality.cennznet.main and epay.nzd.test as payment options

Neeraj-Ghodla:


ncjones:
I would like us to use "cennznet.main" for this asset type as it is more like bitcoin than centrapay.nzd or epay.nzd asset types.

Neeraj-Ghodla:
@ncjones Fixed

ncjones (approved):

PR #79, opened 2021-01-13
Add: New error codes for Cennznet

Neeraj-Ghodla:


heymatthew (approved):

src/transacting.md
```diff
@@ -281,6 +281,9 @@ curl -X POST "https://service.centrapay.com/payments/api/transactions.refund" \
...
+| 195        | 400       | {% break - INVALID_ASSET_ID %}                    | The asset id is invalid                                                                                                                                                                                                                                   |
+| 196        | 400       | {% break - INVALID_WALLET_ADDRESS %}              | The wallet address is invalid                                                                                                                                                                                                                             |
+| 197        | 400       | {% break - INVALID_TRANSACTION %}                 | The transaction is invalid                                                                                                                                                                                                                                |
```
heymatthew, line 286: The problem I have with these descriptions is that they don't add information. The asset id is invalid tells me just as much as INVALID_ASSET_ID

Could we reword this with examples? Is it that the formatting should be a particular way?

Wallet address invalid, does that mean the wallet doesn't exist or can't be found?

Invalid transaction, what kind of things fall into this camp?

Neeraj-Ghodla: What do you think about the new descriptions?

heymatthew (approved):
LGTM

PR #96, opened 2021-03-08
Add update user profile endpoint

daallen303:


src/profile.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+**Optional Fields** (Note at least one field must be provided in the request)
+
+|      Field       |  Type  |                 Description                                                                        |
+| :--------------- | :----- | :------------------------------------------------------------------------------------------------- |
+| givenName        | String | First name                                                                                         |
+| familyName       | String | Last name                                                                                          |
+| featureUpdates   | String | Whether a user would like to receive emails for feature updates (valid values are true or false)   |
+| marketingUpdates | String | Whether a user would like to receive emails for marketing updates (valid values are true or false) |
```
ncjones, line 43: these should be boolean

src/profile.md
```diff
@@ -0,0 +1,56 @@
...
+## Contents
+{:.no_toc .text-delta}
+
+* TOC
+{:toc}
+
+
+## Updating a user's profile
+
+{% endpoint POST https://service.centrapay.com/api/merchants %}
```
ncjones, line 21: wrong endpoint url

src/profile.md
```diff
@@ -0,0 +1,56 @@
+---
+layout: default
+title: Profile
+nav_order: 6
+---
+
+# Profile
+{:.no_toc}
+
+A profile represents a centrapay user's attributes in the cognito user pool.
```
ncjones, line 10: don't mention cognito

daallen303: Removed the cognito reference

src/profile.md
daallen303: Fixed

src/profile.md
daallen303: Updated the docs to be boolean need to update the API, the decision for them to be strings was based on them being persisted as a string in cognito but boolean makes more sense.

src/profile.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Contents
+{:.no_toc .text-delta}
+
+* TOC
+{:toc}
+
+
+## Updating a user's profile
```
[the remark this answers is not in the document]
heymatthew: Otherwise LGTM

heymatthew (approved):

src/profile.md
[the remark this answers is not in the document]
daallen303: Added the experimental tag

PR #100, opened 2021-03-17
Change ruby setup method to use supported project

heymatthew:
Looks like they've archived actions/setup-ruby and the recommendation is
to move to ruby/setup-ruby so that's what this change does.

We have a .ruby-version so it looks like the explicit version isn't
needed. The bundler-cache option just seems like a sensible option.


ncjones (approved):

heymatthew:
Works!

https://github.com/centrapay/centrapay.github.io/runs/2126534758?check_suite_focus=true

PR #101, opened 2021-03-17
Update assetTransfer fields to match our API responses

heymatthew:
We've had additional fields added since this design went out. It only makes sense to update our docs to match.

Based on:

```javascript
class AssetTransfer {
  // ...
  toApiResponse() {
    return {
      id: this.id,
      status: this.status,
      value: this.value,
      assetId: this.sentAssetId,
      assetType: this.assetType,
      description: this.description,
      message: this.message,
      senderName: this.senderName,
      recipientId: this.recipientId,
      recipientAccountId: this.recipientAccountId,
      recipientAlias: this.recipientAlias,
      createdAt: this.createdAt.toISOString(),
      updatedAt: this.updatedAt.toISOString(),
    };
  }
}
```


src/assets/asset-transfers.md
```diff
@@ -118,7 +121,7 @@ curl -X GET "https://service.centrapay.com/api/asset-transfers/M7Kn2stAxNa6ri7h"
   "senderAccountId": "aBc932S9182qwCDqwer",
   "recipientAccountId": "oS3Xom2au3Ooy9aihai",
   "createdAt": "2020-05-01T12:30:00.000Z",
-  "sentAt": "2020-05-01T12:30:00.000Z"
+  "updatedAt": "2020-05-02T01:03:37.222Z"
```
heymatthew, line 124: We never implemented sentAt, but it can be inferred.

We can come back to add this if we need it later.

src/assets/asset-transfers.md
```diff
@@ -94,6 +97,9 @@ The above example has $10 left on a $60 dollar giftcard at the time of transfer.
 
 ## Look up an Asset Transfer
 
+After an asset transfer is resolved we scrub the recipientAlias field to avoid storing PII against
+our model.
```
heymatthew, line 101: Using scrubbed responses in our examples now.

ncjones (approved):

PR #111, opened 2021-03-29
Document Withdrawls and adjust Bank Authority requirements

heymatthew:
We need to make several fields optional in order to be able to support minimal information for withdrawls.

We also need to be able to track this information as it's till required for top ups.

Updates cover bank accounts and also withdraws pages.

### DIFF
--- a/src/fiat/bank-authorities.md
+++ b/src/fiat/bank-authorities.md
@@ -7,10 +7,16 @@ title: Bank Authorities
 # Bank Authorities
 {:.no_toc}
 
-A bank authority represents an individual's consent for Centrapay to transfer
-funds to and from a bank account. In order to move funds between a bank account
-and a Centrapay wallet, there must be a verified bank authority for the
-Centrapay account that the wallet belongs to.
+A bank authority represents an individual's consent for Centrapay to transfer funds to and from a
+bank account.
+
+In order to move funds from a Centrapay wallet to a bank account (withdrawal), we only need the
+minimum required fields. There's no need to authorize direct debit authority, nor verify the account
+using a code on a bank statement.
+
+In order to move funds between a bank account to a Centrapay wallet (top up), we need explicit
+authorization and verification from the user. Bank account authorization and verification status can
+be observed based on the `directDebitAuthorized` and `verified` flags.
 
 ## Contents
 {:.no_toc .text-delta}
@@ -18,26 +24,42 @@ Centrapay account that the wallet belongs to.
 * TOC
 {:toc}
 
+<span id="bank-account-create"></span>
+## Creating a bank account **EXPERIMENTAL**
 
-## Creating a bank authority
+{% endpoint POST https://service.centrapay.com/api/bank-accounts %}
 
-{% endpoint POST https://service.centrapay.com/api/bank-authorities %}
+An example of a minimal POST to create a bank authority.
 
 ```sh
-curl -X POST "https://service.centrapay.com/api/bank-authorities" \
+curl -X POST "https://service.centrapay.com/api/bank-accounts" \
   -H "x-api-key: 1234" \
   -H "content-type: application/json" \
   -d '{
-    "fullName": "John Doe",
     "accountId": "Jaim1Cu1Q55uooxSens6yk",
-    "streetAddress": "11 Shortland St",
-    "suburb": "Auckland Central",
-    "city": "Auckland",
-    "postCode": "1100",
-    "phoneNumber": "+64212345678",
-    "emailAddress": "John.doe@email.com",
     "bankAccountNumber": "12-1234-1234567-123",
     "bankAccountName": "John Doe"
+  }'
+```
+
+An example of a minimal POST to create a bank authority and a direct debit authority.
+
+By including directDebitAuthority, the user accepts our [Direct Debit terms][dd-terms]{:.external}
+and has authority to operate this account.
+
+```sh
+curl -X POST "https://service.centrapay.com/api/bank-accounts" \
+  -H "x-api-key: 1234" \
+  -H "content-type: application/json" \
+  -d '{
+    "accountId": "Jaim1Cu1Q55uooxSens6yk",
+    "bankAccountNumber": "12-1234-1234567-123",
+    "bankAccountName": "John Doe",
+    "directDebitAuthority": {
+      "phoneNumber": "+64212345678",
+      "fullName": "John Doe",
+      "emailAddress": "john.doe@gmail.com"
+    }
   }'
 ```
 
@@ -45,20 +67,20 @@ curl -X POST "https://service.centrapay.com/api/bank-authorities" \
 
 |       Field       |  Type  |                    Description                    |
 | :---------------- | :----- | :------------------------------------------------ |
-| fullName          | String | The first and last name of the user               |
-| phoneNumber       | String | The user's phone number                           |
-| emailAddress      | String | The user's email address                          |
 | bankAccountNumber | String | The user's bank account number                    |
 | bankAccountName   | String | The name on the bank account provided by the user |
 
 **Optional Fields**
 
-|       Field       |  Type  |                    Description                    |
-| :---------------- | :----- | :------------------------------------------------ |
-| streetAddress     | String | The street address of the user                    |
-| suburb            | String | The suburb relating to the user's address         |
-| city              | String | The city relating to the users address            |
-| postCode          | String | The postal code relating to the user's address    |
+Note, fields which have a star (✩) create a direct-debit authority and are required for Top Up. All
+fields below when specified are required together.
+
+|    Field     |  Type   |              Description               |
+| :----------- | :------ | :------------------------------------- |
+| phoneNumber  | String  | ✩ The user's phone number.             |
+| fullName     | String  | ✩ The first and last name of the user. |
+| emailAddress | String  | ✩ The user's email address.            |
+
 
 **Example response payload**
 
@@ -66,15 +88,9 @@ curl -X POST "https://service.centrapay.com/api/bank-authorities" \
 {
   "id": "WRhAxxWpTKb5U7pXyxQjjY",
   "accountId": "Jaim1Cu1Q55uooxSens6yk",
-  "fullName": "John Doe",
-  "streetAddress": "11 Shortland St",
-  "suburb": "Auckland Central",
-  "city": "Auckland",
-  "postCode": "1100",
-  "phoneNumber": "+64212345678",
-  "emailAddress": "John.doe@email.com",
   "bankAccountNumber": "12-1234-1234567-123",
   "bankAccountName": "John Doe",
+  "directDebitAuthorized": true,
   "status": "created",
   "verified": false,
   "createdAt": "2020-06-12T01:17:46.499Z",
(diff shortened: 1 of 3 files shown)


src/fiat/bank-authorities.md
```diff
@@ -7,10 +7,16 @@ title: Bank Authorities
...
+A bank authority represents an individual's consent for Centrapay to transfer funds to and from a
+bank account.
+
+In order to move funds from a Centrapay wallet and a bank account (withdrawl), we only need the
+minimum required fields. There's no need to authorize direct debit authority, nor verify the account
+using a code on a bank statement.
+
+In order to move funds between a bank account and a Centrapay wallet (top up), we need explicit
+authorisation from the user and contact information from them. Before the funds are available, a
+verification code will need to be coppied from a bank statement.
```
maticko, line 19: `copied`

src/fiat/bank-authorities.md
```diff
@@ -7,10 +7,16 @@ title: Bank Authorities
...
-Centrapay account that the wallet belongs to.
+A bank authority represents an individual's consent for Centrapay to transfer funds to and from a
+bank account.
+
+In order to move funds from a Centrapay wallet and a bank account (withdrawl), we only need the
+minimum required fields. There's no need to authorize direct debit authority, nor verify the account
+using a code on a bank statement.
+
+In order to move funds between a bank account and a Centrapay wallet (top up), we need explicit
+authorisation from the user and contact information from them. Before the funds are available, a
```
ncjones, line 18: s/authorisation/authorization

src/fiat/bank-authorities.md
ncjones, line 19: s/coppied/copied

src/fiat/bank-authorities.md
```diff
@@ -30,51 +36,47 @@ curl -X POST "https://service.centrapay.com/api/bank-authorities" \
...
+
+|       Field               |  Type   |                    Description                                                       |
+| :------------------------ | :------ | :----------------------------------------------------------------------------------- |
+| userAuthorizesDirectDebit | Boolean | ✩ The user accepts our Direct Debit terms and has authority to operate this account. |
+| fullName                  | String  | ✩ The first and last name of the user                                                |
+| emailAddress              | String  | ✩ The user's email address                                                           |
+| streetAddress             | String  | Deprecated.                                                                          |
+| suburb                    | String  | Deprecated.                                                                          |
+| city                      | String  | Deprecated.                                                                          |
+| postCode                  | String  | Deprecated.                                                                          |
```
daallen303, line 69: I think we can just remove these

daallen303: the deprecated values

maticko (changes requested):

src/fiat/funds-transfers.md
```diff
@@ -125,3 +126,113 @@ curl -X GET "https://service.centrapay.com/api/topups" \
...
+  "amount": "10000",
+  "status": "created",
+  "createdAt": "2020-05-01T12:30:00.000Z",
+  "updatedAt": "2020-05-01T12:30:00.000Z"
+}
+```
+
+## List withdrawls for authorized accounts **EXPERIMENTAL**
+
+{% endpoint GET https://service.centrapay.com/withdrawls %}
```
ncjones, line 204: /api/withdrawals

src/fiat/bank-authorities.md
```diff
@@ -106,14 +108,39 @@ curl -X GET https://service.centrapay.com/api/bank-authorities/Jaim1Cu1Q55uooxSe
...
+
+## Update information about a bank authority **EXPERIMENTAL**
+
+{% endpoint PUT https://service.centrapay.com/api/bank-authorities/${id} %}
+
+```sh
+curl -X PUT https://service.centrapay.com/api/bank-authorities/Jaim1Cu1Q55uooxSens6yk \
+  -H "x-api-key: 1234" \
+  -H "content-type: application/json" \
+  -d '{ "userAuthorizesDirectDebit": true }'
```
daallen303, line 131: We will need to update the full name and email address too right?

daallen303 (changes requested):

src/fiat/funds-transfers.md
ncjones: I don't want us to make new endpoints that query for all authorized accounts. Can we put the account id as a path param?

src/fiat/bank-authorities.md
heymatthew: I'll add some documentation around the fields, good catch!

src/fiat/funds-transfers.md
ncjones: This API should be forwards compatible with pagination. Can you please use paginated response schema?

{ items: [] }

src/fiat/bank-authorities.md
heymatthew: done 👍

src/fiat/bank-authorities.md
heymatthew: Fixed 👍

heymatthew: 👌 double fixed

src/fiat/funds-transfers.md
heymatthew: Whoops, updated!

src/fiat/bank-authorities.md
heymatthew: They're goneburgers 🍔

src/fiat/bank-authorities.md
heymatthew: Added docs with additional fields 📚

daallen303 (approved):

src/fiat/funds-transfers.md
```diff
@@ -125,3 +126,113 @@ curl -X GET "https://service.centrapay.com/api/topups" \
...
+  "updatedAt": "2020-05-01T12:30:00.000Z"
+}
+```
+
+**Error Responses**
+
+| Status |                     Code                      |                                         Description                                         |
+| :----- | :-------------------------------------------- | :------------------------------------------------------------------------------------------ |
+| 403    | {% break _ BANK_AUTHORITY_WALLET_MISMATCH %}  | The wallet and the bank authority for the top up request do not belong to the same account. |
+| 403    | {% break _ SPENDING_MONTHLY_LIMIT_EXCEEDED %} | The user has gone over their quota for maximum withdawls this month.                        |
```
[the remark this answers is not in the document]
heymatthew: Fair enough, I've gone and removed these docs and have asked @daallen303 to document these states when he comes through and documents top up. We're expecting those in another PR later today :)

src/fiat/funds-transfers.md
heymatthew: > Whoops, updated!

^^ this was in the context of `/api/withdrawls`.

I'll move path into the URL. Should we create endpoints to support topup flows in the same way?

I'll go add  a paginated response.

src/fiat/funds-transfers.md
```diff
@@ -90,7 +91,7 @@ curl -X GET "https://service.centrapay.com/api/topups/WRhAxxWpTKb5U7pXyxQjjY" \
 
 ## List top ups for authorized accounts
 
-{% endpoint GET https://service.centrapay.com/topups %}
+{% endpoint GET https://service.centrapay.com/api/topups %}
```
heymatthew, line 94: Picked up from earlier review by @ncjones

src/fiat/funds-transfers.md
heymatthew: Hey @ncjones , I've gone and updated the response and endpoints to be more account specific. I've also gone and added an EXPERIMENTAL version for topups so we have symmetry.

src/fiat/bank-authorities.md
```diff
@@ -30,51 +36,43 @@ curl -X POST "https://service.centrapay.com/api/bank-authorities" \
...
-| :---------------- | :----- | :------------------------------------------------ |
-| streetAddress     | String | The street address of the user                    |
-| suburb            | String | The suburb relating to the user's address         |
-| city              | String | The city relating to the users address            |
-| postCode          | String | The postal code relating to the user's address    |
+Note, fields which have a star (✩) are required for Top Up.
+
+|       Field               |  Type   |                    Description                                                       |
+| :------------------------ | :------ | :----------------------------------------------------------------------------------- |
+| userAuthorizesDirectDebit | Boolean | ✩ The user accepts our Direct Debit terms and has authority to operate this account. |
```
ncjones, line 63: Should we call this "directDebitAuthorized"? What will the default value be? False makes more sense but is a breaking change.

src/fiat/bank-authorities.md
ncjones: I don't get why this is necessary

src/fiat/bank-authorities.md
heymatthew: Yeah, default should be false.

I was keen on making the subject of this field the user because I wanted to make it pretty clear that calls with this property set were something the account owner had seen and approved of. While we could always point at our docs if there was confusion, making it super clear from the name makes me feel better about scripting.

Apart from making the field shorter, are there other reasons we'd change this name?

It feels like we should link to our DD terms form here too.

src/fiat/funds-transfers.md
```diff
@@ -57,13 +57,14 @@ curl -X POST "https://service.centrapay.com/api/topups" \
...
+| 403    | {% break _ BANK_AUTHORITY_WALLET_MISMATCH %}     | The wallet and the bank authority for the top up request do not belong to the same account.                                                                                                                                          |
+| 403    | {% break _ MAX_INFLIGHT_TOPUPS_EXCEEDED %}       | The bank authority already has ten pending top ups, which is the maximum a bank authority can have at any one time.                                                                                                                  |
 | 403    | {% break _ MAX_INFLIGHT_TOPUP_AMOUNT_EXCEEDED %} | The top up can not be created because it is above the max amount of funds a bank authority can have pending at any one time. The max amount is $1000.00 for verified bank authorities and $100.00 for non-verified bank authorities. |
+| 403    | {% break _ MISSING_BANK_AUTHORITY_INFORMATION %} | Bank authority is missing required fields for topup. See [bank authorities](bank-authorities.html#bank-fields-anchor) documentation for a list of required fields.                                                                   |
```
ncjones, line 65: I think this should be DIRECT_DEBIT_NOT_AUTHORIZED. A bank account should be authorized or not for DD and not some state in between. Every time DD auth is added to a bank account then the identifying params should be provided.

heymatthew: Are you suggesting that our update endpoint should force all 3 fields to be submitted together?

src/fiat/bank-authorities.md
ncjones: I don't like it. At the moment user authority is implied. I don't think there needs to be extra clarity.

heymatthew: I've updated this to `directDebitAuthorized`

src/fiat/funds-transfers.md
heymatthew: Our intention is to roll out this:

Add a new bank account from withdrawl

The problem with this screen is that it lacks `userAuthorizesDirectDebit`, `fullName` and `emailAddress` and all 3 optional fields would be required for topup.

Calling this `DIRECT_DEBIT_NOT_AUTHORIZED` implies you might be able to do this without fullName` and `emailAddress` filled out.

I'd prefer to keep with `MISSING_BANK_AUTHORITY_INFORMATION` to nod to all optional fields and make our integrations check against `topupFieldsRequired`.

maticko: Other required values are picked from Cognito User Profile. UI could check if those values are present and ask for them if they are not.

heymatthew: I've talked this over with @maticko, and here are the options we've explored:

1. Breaking change
    - force `DIRECT_DEBIT_NOT_AUTHORIZED`
        - have 3 fields you have to specify together
        - this one is good, but we're not keen on making a non-breaking change
    - keep `MISSING_BANK_AUTHORITY_INFORMATION`
        - allow any field to be updated anytime (but do it together in our UX)
        - have 3 fields you can update anytime till they're verified
        - this won't work with our S3 plan, so it's a no go

2. Non breaking change
    - create a new endpoint and deprectate the old
      i.e. https://service.centrapay.com/api/accounts/${accountId}/bank-authorities
    - force `DIRECT_DEBIT_NOT_AUTHORIZED`
        - have 3 fields you have to specify together

We're feeling pretty good about number 2 and I'll update the docs to reflect this. If you can think of another angle or a different way, please sing out.

heymatthew: This is now updated such that

- We have added a DEPRECATED flag to existing create endpoint
- We have a new API endpoint for creating bank authorities
- We have a new API endpoint for finalizing a bank authority
- PII is removed from all bank authority responses
- DIRECT_DEBIT_NOT_AUTHORIZED is thrown from topup create endpoint
- directDebitAuthorized has been added to all API responses
- topupFieldsRequired has been removed from all API responses

I believe this covers our requirements. Please have another review and tell me what-cha think :)

src/fiat/bank-authorities.md
```diff
@@ -83,43 +76,114 @@ curl -X POST "https://service.centrapay.com/api/bank-authorities" \
...
+}
+```
+
+<span id="finalize-anchor"></span>
+## Finalize information about a bank authority **EXPERIMENTAL**
+
+{% endpoint PUT https://service.centrapay.com/api/${accountId}/bank-authorities/${id}/finalize %}
+
+```sh
+curl -X PUT https://service.centrapay.com/api/accounts/Jaim1Cu1Q55uooxSens6yk/bank-authorities/WRhAxxWpTKb5U7pXyxQjjY/finalize \
```
ncjones, line 120: 1. Finalize is too general... what if we need some other state change later?
2. Account id path param is redundant... bank authority id is globally unique
3. We follow [rest without put][1] as a design principle so usually we a posting to a sub resource.

What about `POST https://service.centrapay.com/api/bank-authorities/${id}/authorize-direct-debit`?

[1]: https://www.thoughtworks.com/radar/techniques/rest-without-put

src/fiat/bank-authorities.md
```diff
@@ -7,35 +7,34 @@ title: Bank Authorities
...
 {:.no_toc}
 
-A bank authority represents an individual's consent for Centrapay to transfer
-funds to and from a bank account. In order to move funds between a bank account
-and a Centrapay wallet, there must be a verified bank authority for the
-Centrapay account that the wallet belongs to.
+A bank authority represents an individual's consent for Centrapay to transfer funds to and from a
+bank account.
+
+In order to move funds from a Centrapay wallet and a bank account (withdrawl), we only need the
```
Ray2126, line 13: from a Centrapay wallet **to** a bank account

src/fiat/funds-transfers.md
```diff
@@ -126,3 +127,113 @@ curl -X GET "https://service.centrapay.com/api/topups" \
   }
 ]
 ```
+
+## Creating a Withdrawl **EXPERIMENTAL**
```
Ray2126, line 131: Should this be **Withdrawal** instead of Withdrawl

heymatthew: Yep, bit of a typo. There's a fix in the pipeline coming!

src/fiat/bank-authorities.md
heymatthew: That does read better, a fix is in the pipe.

src/fiat/bank-authorities.md
heymatthew: Agreed 💖, I've gone and updated this and pushed it up.

src/auth.md
```diff
@@ -143,6 +143,7 @@ This example demonstrates how to integrate a browser based application with the
 Centrapay OAuth server using the [oidc-client-js][oidcjs]{:.external}
 JavaScript library.
 
+<span id="oidcjs-example">
```
daallen303, line 146: Should this span be closed?

daallen303 (approved):

src/auth.md
heymatthew: Yes

heymatthew: Looks like this was added in error. There were two IDs on the page which wasn't so great.

I've removed the span so we should be golden 🕺

src/fiat/bank-authorities.md
```diff
@@ -83,47 +98,166 @@ curl -X POST "https://service.centrapay.com/api/bank-authorities" \
...
 
-{% endpoint GET https://service.centrapay.com/api/bank-authorities/${id} %}
+{% endpoint POST https://service.centrapay.com/bank-accounts/${bankAccountId}/direct-debit-authorities %}
+
+```sh
+curl -X PUT https://service.centrapay.com/api/bank-accounts/WRhAxxWpTKb5U7pXyxQjjY/direct-debit-authorities \
+  -H "x-api-key: 1234" \
+  -H "content-type: application/json" \
+  -d '{
+    "authorized": true,
```
Neeraj-Ghodla, line 119: Should this also contain `accountId` as we used `accountId` to asset permissions?

src/fiat/bank-authorities.md
```diff
@@ -83,47 +98,166 @@ curl -X POST "https://service.centrapay.com/api/bank-authorities" \
...
+| authorized   | Boolean | ✩ The user accepts our [Direct Debit terms][dd-terms]{:.external} and has authority to operate this account. |
+| phoneNumber  | String  | ✩ The user's phone number.                                                                                   |
+| fullName     | String  | ✩ The first and last name of the user.                                                                       |
+| emailAddress | String  | ✩ The user's email address.                                                                                  |
+
+**Error Responses**
+
+| Status |                           Code         |                          Description                               |
+| :----- | :------------------------------------- | :----------------------------------------------------------------- |
+| 403    | {% break _ BANK_AUTHORITY_FINALIZED %} | This bank authority cannot be changed as all fields have been set. |
```
Neeraj-Ghodla, line 160: Can we change the error code to something like `BANK_AUTHORITY_ALREADY_DIRECT_DEBIT_AUTHORIZED`?

src/fiat/bank-authorities.md
```diff
@@ -83,47 +98,166 @@ curl -X POST "https://service.centrapay.com/api/bank-authorities" \
...
+| :----- | :------------------------------------------------------- | :--------------------------------------------------------------------------------------------------------- |
+| 403    | {% break _ BANK_AUTHORITY_LIMIT_EXCEEDED %}              | The account already has the max amount of bank authorities.                                                |
 | 403    | {% break _ BANK_AUTHORITIES_FOR_BANK_ACCOUNT_EXCEEDED %} | There are already two bank authorities for the provided bank account number, which is the maximum allowed. |
 
-## Get information about a bank authority
+<span id="direct-debit-authority"></span>
+## Adding a direct debit authority to a bank account **EXPERIMENTAL**
 
-{% endpoint GET https://service.centrapay.com/api/bank-authorities/${id} %}
+{% endpoint POST https://service.centrapay.com/bank-accounts/${bankAccountId}/direct-debit-authorities %}
```
Ray2126, line 112: Is this missing /api

src/fiat/bank-authorities.md
heymatthew: Updated this :)

src/fiat/bank-authorities.md
heymatthew: Fixed 💖

src/fiat/bank-authorities.md
```diff
@@ -7,74 +7,89 @@ title: Bank Authorities
...
-| streetAddress     | String | The street address of the user                    |
-| suburb            | String | The suburb relating to the user's address         |
-| city              | String | The city relating to the users address            |
-| postCode          | String | The postal code relating to the user's address    |
+Note, fields which have a star (✩) create a direct-debit authority and are required for Top Up. All
+fields below when specified are required together.
+
+|    Field     |  Type   |                    Description                                                                               |
+| :----------- | :------ | :----------------------------------------------------------------------------------------------------------- |
+| authorized   | Boolean | ✩ The user accepts our [Direct Debit terms][dd-terms]{:.external} and has authority to operate this account. |
```
heymatthew, line 78: remove this and we'll have symmetrical docs

src/fiat/bank-authorities.md
ncjones: @heymatthew we could shorten it to DIRECT_DEBIT_ALREADY_AUTHORIZED

maticko (approved):

src/fiat/bank-authorities.md
heymatthew: Done.

src/fiat/bank-authorities.md
heymatthew: Done.

PR #112, opened 2021-03-30
Decouple anchors from titles

heymatthew:
Fixes anchors on OIDC flow page, and decouples anchors in other places too.

### DIFF
--- a/src/auth.md
+++ b/src/auth.md
@@ -35,7 +35,7 @@ curl -X GET "https://service.centrapay.com/api/account-memberships" -H "Authoriz
 
 API keys provide enduring access to a single Centrapay account.
 
-See: [Creating API Keys]({% link accounts.md %}#creating-an-api-key).
+See: [Creating API Keys]({% link accounts.md %}#creating-api-key).
 
 The Centrapay test merchant API key is available to test creating payment requests:
 `f32c5497297084e5354b47c40d5ccacb109ce483`.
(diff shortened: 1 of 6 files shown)


src/auth.md
```diff
@@ -63,7 +63,7 @@ website (such as "https://yourapp.example.com/oidc-callback") or mobile app
 
 Your application can use any OIDC client to negotiate the authentication flow
 but it must support OIDC authorization code flow with [PKCE][pkce]{:.external}.
-See the [Example Web Browser OIDC Consumer][oidcjs-example] section below for a
+See the [Example Web Browser OIDC Consumer](#oidcjs-example-anchor) section below for a
```
heymatthew, line 66: This link is currently broken in the wild.

ncjones: I don't like the anchor suffix. It's redundant.

heymatthew: Using a suffix that'll never show up in a title ensures we wont end up generating multiple IDs on the page. The thinking is that someone is unlikely to put the word 'anchor' in their heading, but it's likely that a heading may change without updating the dom.

I've gone and removed the `-anchor` suffix. Lets just try to keep the anchor IDs different to what we generate with titles and it should all be fine 👌

ncjones: We should auto generate anchors from section headings and the linter should detect broken links.

ncjones (approved):

src/auth.md
heymatthew: +1, lets investigate a linter in a followup PR and remove this when we get there.

PR #132, opened 2021-04-21
WIP Create and send

heymatthew:
Previously reverted by #131

Reinstates #122


heymatthew:
This PR is suffering from bit rot so I'm closing it.

PR #135, opened 2021-04-23
document asset transfer lifecycle

maticko:


ncjones:
I think we should not mention indexes in the public docs.

ncjones (approved):

heymatthew (approved):

PR #141, opened 2021-05-12
Host api docs under /api path

ncjones:


heymatthew (approved):
LGTM, just whitespace :)

PR #144, opened 2021-05-15
Assets cleanup

ncjones:


src/assets/assets.md
```diff
@@ -12,8 +12,14 @@ redirect_from:
...
-Assets are resources that can be used to satisfy payment requests. Assets are
-owned by an account and can be sent to other users.
+Centrapay digital assets are resources that represent the ability for a
+Centrapay account to perform transactions where value is exchanged. Assets can
+be spent to satisfy [Payment Requests][], withdrawn to a bank account via
+[Funds Transfers][] and sent to other Centrapay users via [Asset Transfers][].
+
+Assets are categorized as either Money, Gift Cards or Tokens. Depending on its
+asset category, an asset will have different attributes available and different
+rules governing how it can be obtained, shared or spent.
```
heymatthew, line 22: 👍

src/assets/assets.md
```diff
@@ -22,6 +28,98 @@ owned by an account and can be sent to other users.
 * TOC
 {:toc}
 
+## Asset Model
+
+All assets have the following fields along with the additional fields that are
+specific to its category.
+
+**Required Fields**
```
heymatthew, line 36: It feels a little strange to call these "required". On other pages in our APIs, this term sets out an expectation on our API consumer to provide those fields. However, this is instead a requirement on our API response.

Could we add words or use a different title here?

Maybe something like "Guaranteed response fields"?

src/assets/assets.md
```diff
@@ -22,6 +28,98 @@ owned by an account and can be sent to other users.
...
+
+| Field          | Type           | Description                                                          |
+|:---------------|:---------------|:---------------------------------------------------------------------|
+| issuer         | String         | The identifier for the issuer of the gift card.                      |
+| currency       | String         | Currency code, eg "NZD"                                              |
+| balance        | Numeric String | Current balance in the currency's smallest denomination (ie. cents). |
+| initialBalance | Numeric String | The balance when the asset was created.                              |
+
+**Optional Fields**
```
heymatthew, line 84: Ditto, it'd be nice to have different terms for returned fields.

How about "Optional response fields"

src/assets/external-assets.md
```diff
@@ -2,19 +2,17 @@
 layout: default
 grand_parent: API Reference
 parent: Assets
-title: Discrete Assets
-permalink: /api/discrete-assets
+title: External Assets
+permalink: /api/external-assets
 redirect_from:
   - /assets/discrete-assets
+  - /api/discrete-assets
```
heymatthew, line 9: Could we also redirect from `/api/asset-categories` here?

```suggestion
  - /api/discrete-assets
  - /assets/asset-categories
```

heymatthew (changes requested):
Just a few things

heymatthew (approved):

PR #154, opened 2021-06-03
Add documentation for listing merchants of an account

IsaacMurtagh:


src/api/accounts.md
```diff
@@ -309,3 +309,44 @@ curl -X GET "https://service.centrapay.com/api/users/1234/account-memberships" \
   }
 ]
 ```
+
+## List merchants for account
+
+List merchants for authorized accounts:
```
ncjones, line 315: This line can be removed. The "for authorized accounts" pattern does not apply here and is deprecated. It was used to simplify listing resources when a front-end application does not know which accounts the user has access to. We want clients to always specify the Centrapay account id  from now on.

IsaacMurtagh: Okay

ncjones (approved):

PR #185, opened 2021-06-28
Document optional Payment Option fields on payment request

heymatthew:


src/api/payment-requests/payment-requests.md
```diff
@@ -77,6 +77,18 @@ version (documented on this page) and the "legacy" version (documented at
 | assetType | String             | An [Asset][] type reference.                                            |
 | amount    | {% dt BigNumber %} | The value required to pay using the canonical units for the asset type. |
 
+{% h4 Optional Fields %}
+
+| Field                | Type   | Description                                                               |
+|----------------------|--------|---------------------------------------------------------------------------|
+| bitcoinPublicAddress | String | ★  Custodian account where we accept and settle Bitcoin from.             |
+| cennzPublicAddress   | String | ★  Custodian account where we accept and settle CENNZ from.               |
+| wavesPublicAddress   | String | ★  The merchant's Waves address to send Zap tokens to.                    |
```
ncjones, line 86: * "wavesPublicAddress" was chosen for Zap config because it seemed like the most idiomatic name. I don't know if the same is true for cennz and bitcoin. Maybe "bitcoinAddress" or "bitcoinWallet". Not sure about CENNZ.
* Whether these are custodian or merchant addresses is not relevant here. That detail can go on the merchant config docs. Here the caller just needs to know that a payment should be made to that address.

ncjones (approved):

PR #197, opened 2021-07-06
Adjust payment request examples

Ray2126:


ncjones:
@Ray2126 can you restore the deleted lines from the full line item example?

ncjones (approved):

PR #235, opened 2021-08-23
Add settlement config model to merchants

IsaacMurtagh:


src/api/merchants/merchants.md
```diff
@@ -72,6 +79,12 @@ which define the payment methods available for a Payment Request.
 }
 ```
 
+{% h4 Error Responses %}
+
+| Status |                  Code                   |                    Description                    |
+| :----- | :-------------------------------------- | :------------------------------------------------ |
+| 403    | {% break _ INVALID_SETTLEMENT_CONFIG %} | One or more settlement config options are invalid |
```
ncjones, line 86: I think we should have a specific error message for each rule that is violated.

ncjones: Elsewhere we used "ACCOUNT_MISMATCH"

ncjones: Should we us "BANK_ACCOUNT_MISMATCH"?

ncjones: BTW, if the target bank account is not found or not authorized it will be 403 FORBIDDEN. This error is only when the bank account is authorized but owned by a different account.

IsaacMurtagh: Yep, I'll update this

IsaacMurtagh: On the endpoint GET /api/bank-accounts/{bankAccountId}, we don't expose if an account does not exist. We just return forbidden if they don't have access to it. Should we do the same?

ncjones (approved):

PR #244, opened 2021-09-09
Payment activities paginated

heymatthew:


src/api/payment-requests/payment-activities.md
```diff
@@ -77,57 +76,67 @@ descending activity created date.
...
-    "merchantId": "5ee0c486308f590260d9a07f",
-    "merchantAccountId": "C4QnjXvj8At6SMsEN4LRi9",
-    "merchantConfigId": "5ee168e8597be5002af7b454",
-    "createdAt": "2021-06-12T01:15:46.000Z",
-    "createdBy": "crn::user:0af834c8-1110-11ec-9072-3e22fb52e878",
-    "paymentRequestCreatedBy": "crn::user:0af834c8-1110-11ec-9072-3e22fb52e878"
-  }
-]
+{
+  "nextPageKey": "5ee0c486308f590260d9a07f#2021-06-12T01:15:46.000Z",
```
heymatthew, line 89: We've reviewed the way that [Dynamo pages define page tokens](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#API_Query_RequestSyntax) and it looks like it's the SK and PK. Here's an example of an exclusiveStartKey used to page through wallets in a test database:

```json
        "exclusiveStartKey": {
          "sk": {
            "S": "Wallet#5gXCavNUKHdYi1FB9XkcE1"
          },
          "pk": {
            "S": "Account#1vv74bzenjgpvscg4uug9565s47bw0"
          }
        }
```

However, in the context of the listing endpoint we should be able to construct a dynamo document out of a minimal set of identifiers. This will require some engineering to translate this on the other end.

For this endpoint, the partition key `"5ee0c486308f590260d9a07f#2021-06-12T01:15:46.000Z` is based on the paymentRequestId and sort key is based on createdAt.

This doesn't matter to an API consumer, but it's a realistic key based on our example.

src/api/assets/wallets.md
```diff
@@ -117,6 +117,8 @@ send money from this Wallet.
 
 ## Listing Wallet Transactions **EXPERIMENTAL**
 
+Returns a [paginated][] list of Wallet Transactions.
```
heymatthew, line 120: @ncjones , we should chat about linking this in from all these other pages.

ncjones (approved):

PR #246, opened 2021-09-10
Format all JSON examples

ncjones:
The new "json" liquid tag accepts YAML as body content and outputs
formatted and highlighted JSON.

### DIFF
--- a/src/_plugins/json.rb
+++ b/src/_plugins/json.rb
@@ -1,0 +1,36 @@
+require 'yaml'
+require 'json'
+
+module Jekyll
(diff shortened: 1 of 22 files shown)


ncjones:
@heymatthew  This guarantees syntactically valid examples with consistent formatting. Optionally it allows us to use simpler yaml syntax to define the examples. Most of the examples are left as JSON but can be simplified on an on-going case-by-case basis.

heymatthew (commented):
Nice job!

heymatthew (approved):

PR #260, opened 2021-09-30
Update Activity Types to include expired, cancelled and declined

heymatthew:


src/api/payment-requests/payment-activities.md
```diff
@@ -49,11 +49,14 @@ Payment Activities are created when a Payment Request has been **created**, **pa
...
-| payment | Payment Request was paid.     |
-| refund  | Payment Request was refunded. |
+| Name      | Description                                        |
+|-----------|----------------------------------------------------|
+| request   | [Payment Request][] was created.                   |
+| cancelled | [Payment Request][] was cancelled by the merchant. |
+| expired   | [Payment Request][] wasn't paid before time out.   |
+| declined  | [Payment Request][] was declined by the shopper.   |
+| payment   | [Payment Request][] was paid.                      |
+| refund    | Funds were returned to the shopper.                |
```
ncjones, line 59: Should this be: request, payment, refund, cancellation, expiry

heymatthew: I think you've got 2 pieces of feedback here, one is the ordering and the other is the text in the type.

The ordering we can update no trouble.

When talking about activity, I expect language that implies actions that happened in the past. This is in contrast to your suggestion, cancellation and expiry describe a process that's happening, where as expired and cancelled describe activities that have happened.

Could you give me some more context? Come grab me.

heymatthew: This has been updated.

daallen303 (approved):

PR #263, opened 2021-10-05
Add get default merchant config endpoint to merchant config docs

IsaacMurtagh:


IsaacMurtagh:
Closed as we don't want this endpoint now

PR #264, opened 2021-10-05
Add externalId to account model and createAccount route

Ray2126:


src/api/accounts/accounts.md
```diff
@@ -32,17 +32,17 @@ individual account.
...
+| id         | String             | The unique identifier.                                                             |
+| type       | String             | Account type, must be either 'org' or 'individual'.                                |
+| name       | String             | The display name of the Account.                                                   |
+| test       | Boolean            | A flag which is only present if the account is for testing.                        |
+| createdAt  | {% dt Timestamp %} | When the Account was created.                                                      |
+| modifiedAt | {% dt Timestamp %} | When the Account was updated.                                                      |
+| createdBy  | {% dt CRN %}       | The User or API Key that created the Account.                                      |
+| modifiedBy | {% dt CRN %}       | The User or API Key that updated the Account.                                      |
+| externalId | String             | Id of an external account that this Centrapay Account is mapped to; if applicable. |
```
IsaacMurtagh, line 45: If applicable. Does that mean this externalId is sometimes there?
If so, can we split the Account model into requiredFields and optionalFields.
optional fields would be, test and externalId

IsaacMurtagh (changes requested):

src/api/accounts/accounts.md
Ray2126: Yep, I've split this out now.

IsaacMurtagh (approved):

PR #270, opened 2021-10-18
Remove accountId requirement from Integration Request

MeganSteenkamp:
Remove accountId as required from Integration Request model, and remove from create integration request endpoint.
It is not required as this can be deduced from the merchantId.


src/api/integration-requests.md
```diff
@@ -25,18 +25,18 @@ Integration Requests enable terminal integrations and acceptance of asset types
 |   Field    |  Type  |                                                           Description                                                            |
 | :--------- | :----- | :------------------------------------------------------------------------------------------------------------------------------- |
 | id         | String | The Integration Request's unique identifier.                                                                                     |
-| accountId  | String | The Integration Request's owning Centrapay Account id.                                                                           |
```
IsaacMurtagh, line 28: AccountId is still a mandatory field of an integration request. This PR should only remove the accountId coming from the request body

MeganSteenkamp: I have reverted the model to how it was

ncjones (approved):

PR #277, opened 2021-11-05
Change authorization parameter value for the test uplink

daallen303:
I think None implies you don't supply an authorization param, maybe any makes sense?


heymatthew (approved):
LGTM, tag @ncjones

PR #283, opened 2021-11-17
Use plural for media uploads, add get media content endpoint

MeganSteenkamp:


ncjones:
The response location might not be an image

ncjones (approved):

PR #295, opened 2021-12-01
Adding documentation for paying and refunding payment requests

laurencemcknight:
Adding documentation for proposed API endpoints used to pay and refund payment requests.
The endpoints are currently only for single asset payment requests, multi asset payment requests will come in future.
These endpoints are needed to allow merchants to refund payments from the payment request view.

### DIFF
--- a/src/api/payment-requests/payment-requests.md
+++ b/src/api/payment-requests/payment-requests.md
@@ -467,12 +467,148 @@ them to find the Payment Request and proceed to pay.
 {% endjson %}
 
 
+### Pay a Payment Request by Id **EXPERIMENTAL**
+
+There are two methods of paying a payment request.
+The first uses Centrapay [Assets] and requires you to provide the Id and the type of the asset.
+Alternatively you can provide an external transaction Id and the Centrapay [Asset Type][] for any payments that we support. An example of an external transaction would be a Bitcoin payment.
+
+{% reqspec %}
+  POST '/api/payment-requests/{paymentRequestId}/pay'
+  auth 'jwt'
+  example {
+    title 'Pay a Payment Request with a Centrapay asset'
+    body ({
+      "assetType": "centrapay.nzd.main",
+      "assetId": "WRhAxxWpTKb5U7pXyxQjjY"
+    })
+  }
+
+  example {
+    title 'Pay a Payment Request using external transaction'
+    body ({
+      "assetType": "bitcoin.main",
+      "transactionId": "VMXMkUttDGTVz4ESv5ND56",
+    })
+  }
+
+{% endreqspec %}
+
+{% h4 Example response payload %}
+
+{% json %}
+{
+  "id": "MhocUmpxxmgdHjr7DgKoKw",
+  "url": "https://app.centrapay.com/pay/MhocUmpxxmgdHjr7DgKoKw",
+  "merchantId": "26d3Cp3rJmbMHnuNJmks2N",
+  "merchantName": "Centrapay Café",
+  "configId": "5efbe2fb96c08357bb2b9242",
+  "value": {
+    "currency": "NZD",
+    "amount": "1000"
+  },
+  "paymentOptions": [
+    {
+      "amount": "1000",
+      "assetType": "centrapay.nzd.main"
+    },
+    {
+      "amount": "1000",
+      "assetType": "cca.coke.main"
+    }
+  ],
+  "status": "paid",
+  "createdAt": "2021-06-08T04:04:27.426Z",
+  "updatedAt": "2021-06-08T04:04:27.426Z",
+  "expiresAt": "2021-06-08T04:06:27.426Z",
+  "liveness": "main",
+  "expirySeconds": 120,
+  "paidBy": {
+    "assetTotals": [
+      {
+        "type": "centrapay.nzd.main",
+        "description": "Centrapay NZD",
+        "settlementDate": "2021-06-28T04:04:27.426Z",
+        "total": {
+          "amount": "1000",
+          "currency": "NZD"
+        }
+      }
+    ]
+  }
+}
+{% endjson %}
+
+### Refund a Payment Request by Id **EXPERIMENTAL**
+
+{% reqspec %}
+  POST '/api/payment-requests/{paymentRequestId}/refund'
+  auth 'jwt'
+  example {
+    title 'Refund a Payment Request'
+    body ({
+      "value": {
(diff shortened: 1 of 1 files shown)


src/api/payment-requests/payment-requests.md
```diff
@@ -467,12 +467,168 @@ them to find the Payment Request and proceed to pay.
 {% endjson %}
 
 
+### Pay a Payment Request by Id
```
tom-blake, line 470: Do we want to mark these as experimental as there are some unresolved Qs around multi-asset refunds etc?

src/api/payment-requests/payment-requests.md
```diff
@@ -467,12 +467,168 @@ them to find the Payment Request and proceed to pay.
...
+
+There are two methods of paying a payment request.
+The first uses Centrapay [Assets] and requires you to provide the Id of the asset to use for payment.
+Alternatively you can provide a transaction hash and the Centrapay [Asset Type] for any crypto payments that we support.
+
+{% reqspec %}
+  POST '/api/payment-requests/{paymentRequestId}/payment'
```
tom-blake, line 477: Suggestion to update `/payment` to `/pay` to match the action/verb being performed by the user - thoughts?

Looks like `/pay` is used for our legacy routes too which is why it stood out to me

src/api/payment-requests/payment-requests.md
```diff
@@ -467,12 +467,168 @@ them to find the Payment Request and proceed to pay.
...
+    {
+      "amount": "1000",
+      "assetType": "centrapay.nzd.main"
+    },
+    {
+      "amount": "1000",
+      "assetType": "cca.coke.main"
+    }
+  ],
+  "status": "paid",
```
tom-blake, line 582: Copy-pasta error here? Should this be `refunded` (or whatever we set it to)?

src/api/payment-requests/payment-requests.md
```diff
@@ -467,12 +467,168 @@ them to find the Payment Request and proceed to pay.
...
+  POST '/api/payment-requests/{paymentRequestId}/refund'
+  auth 'jwt'
+  example {
+    title 'Refund a Payment Request'
+    body ({
+      "value": {
+        "amount": "100",
+        "currency": "NZD",
+      },
+      "externalReference": "2021-12-01T01:31:46+00:00",
```
tom-blake, line 553: This guy surprised me - I had no idea what it meant. A couple of questions:
- The [legacy docs](https://docs.centrapay.com/api/legacy-payment-requests#refunding-a-transaction) describe what a refund with an external reference is. Do we want a similar section here?
- What about the option to refund without an external reference? Have we considered supporting that / do we need to?

tom-blake (changes requested):

src/api/payment-requests/payment-requests.md
ncjones: Yes I think so. Overall this looks nice and simple. There's a bunch of context you might be missing so I've added notes to the internal design.

src/api/payment-requests/payment-requests.md
daallen303: This is something we will need to discuss. This idea of paid being a final status has been around with the legacy payment flow. If we add other statuses a fully refunded payment request can just be refunded but partially refunded payments I'm not sure what it would be.

src/api/payment-requests/payment-requests.md
daallen303: Sorry, it is an optional field, we should add documentation that explains that. The original need for the external reference in order to do partial refunds isn't going to be needed anymore.

src/api/payment-requests/payment-requests.md
daallen303: Yeah, agreed. We will add the experimental flag to it.

src/api/payment-requests/payment-requests.md
daallen303: The idea here is that a payment transaction is created when this endpoint is called. In the future, you might post multiple times to this endpoint to satisfy a payment request.  Happy to discuss though.

daallen303: This is updated to /pay and /refund

src/api/payment-requests/payment-requests.md
daallen303: Nathan and I decided to keep the status as paid, so that we aren't adding any complexity to our API contract.

src/api/payment-requests/payment-requests.md
```diff
@@ -467,12 +467,148 @@ them to find the Payment Request and proceed to pay.
...
+  POST '/api/payment-requests/{paymentRequestId}/refund'
+  auth 'jwt'
+  example {
+    title 'Refund a Payment Request'
+    body ({
+      "value": {
+        "amount": "100",
+        "currency": "NZD",
+      },
+      "externalRef": "e8df06e2-13a5-48b4-b670-3fd6d815fe0a",
```
tom-blake, line 554: Looks like my original comment has been hidden after the changes to this file - just a reminder to add some documentation on this field :)

Also looks like it's been renamed from `externalReference` to `externalRef` -> I'd prefer the non-abbreviated `externalReference` as humans read these docs 👍

src/api/payment-requests/payment-requests.md
tom-blake: Yeah sure thing - I'm happy with that now. Will revisit once we get further into partial payments I'm sure, it seems like something that would be useful to enable consumers to track themselves

tom-blake (approved):
@laurencemcknight @daallen303 Nice work guys! Just add some docs in about `externalReference` and I reckon it's good to go

I've approved to unblock you in the meantime 💪

src/api/payment-requests/payment-requests.md
laurencemcknight: We renamed `externalReference` to `externalRef` to provide consistency with the variable names used when creating a payment request (https://docs.centrapay.com/api/payment-requests#create-a-payment-request) but I will talk to @daallen303 tomorrow about the naming because I agree that `externalReference` would be more understandable.
We didn't feel the need to add additional documentation as documentation for `externalRef` already exists on this page but please let me know if you think it needs to expanded upon. 🤠

daallen303: Yeah pretty much what Laurence said. If we change it for this endpoint we should also change it for the others.

PR #300, opened 2021-12-15
Adding error code for attempting to pay with an invalid asset type

laurencemcknight:
The new payment endpoint will throw a INVALID_ASSET_TYPE error if the user attempts to pay with an invalid asset type. The new endpoint will not accept Pocket Vouchers and will throw this error if Pocket Vouchers are used as the payment asset type.

### DIFF
--- a/src/api/payment-requests/payment-requests.md
+++ b/src/api/payment-requests/payment-requests.md
@@ -539,6 +539,12 @@ Alternatively you can provide an external transaction Id and the Centrapay [Asse
 }
 {% endjson %}
 
+{% h4 Error Responses %}
+
+| Status | Code                    | Description                                                                                                                                                                                 |
+|:-------|:------------------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


src/api/payment-requests/payment-requests.md
```diff
@@ -539,6 +539,12 @@ Alternatively you can provide an external transaction Id and the Centrapay [Asse
 }
 {% endjson %}
 
+{% h4 Error Responses %}
+
+| Status | Code                    | Description                                                                                                                                              |
+|:-------|:------------------------|:---------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 403    | INVALID_ASSET_TYPE      | Either the merchant is not configured with the provided asset type, or the asset type can not be used as it is only supported by our [Legacy Payment API][]. |
```
Ray2126, line 546: There's another case where the asset type just doesn't exist.

laurencemcknight: Good point, I've just added that in now.

tom-blake (approved):

PR #306, opened 2022-01-12
Define max length

MeganSteenkamp:
Create plugin to define max length.
Add length limits to asset transfer page.

Plugin will display text '{num} character limit.'


GeekOnGadgets (approved):
LGTM

PR #311, opened 2022-01-21
Update the refund response to be a refund payment activity

daallen303:


src/api/payment-requests/payment-requests.md
```diff
@@ -567,45 +567,20 @@ Alternatively you can provide an external transaction Id and the Centrapay [Asse
...
-    ],
-  }
-}
+  "merchantId": "5ee0c486308f590260d9a07f",
+  "merchantAccountId": "C4QnjXvj8At6SMsEN4LRi9",
+  "merchantConfigId": "5ee168e8597be5002af7b454",
+  "createdAt": "2021-06-12T01:17:00.000Z",
+  "createdBy": "crn::user:0af834c8-1110-11ec-9072-3e22fb52e878",
+  "paymentRequestCreatedBy": "crn::user:0af834c8-1110-11ec-9072-3e22fb52e878",
+  "activityNumber": "3",
```
heymatthew, line 581: This exposes a lot about our modelling. Could we rename this to something a bit more friendly to integrators.

I suggest `version` would be good here.

heymatthew: Could we also record this in notion?

daallen303: This response matches the response from listing payment activities e.g. https://docs.centrapay.com/api/payment-activities#list-payment-activities-for-payment-request-experimental

daallen303: I think it would be best to be consistent with those API's.

heymatthew: This is not a conversation we can have on a PR. Could you pleases update the document in notion so we can have this conversation there?

These are all labelled draft and experimental so we should be able to change them.

Discussion about version vs activityNumber was made in ADR 66 and will be depended on by 3rd parties. Lets talk this out.

daallen303: 👍

heymatthew (approved):

PR #323, opened 2022-02-11
Update balance to availableBalance for asset APIs

daallen303:
https://www.notion.so/centrapay/65a900a58a64424e85e3ccbaad674f51?v=f3704058540e48bc8c3680e61c76b2e3&p=755fdcd9420b44f2a5001d6a5d7728ee
Based on this story it seems like we are going to be returning available balance in all of our asset API's. I updated the docs to reflect that but will wait on confirmation before making the changes.

### DIFF
--- a/src/api/assets/assets.md
+++ b/src/api/assets/assets.md
@@ -40,16 +40,17 @@ specific to its category.
 
 {% h4 Required Fields %}
 
-| Field       | Type               | Description                                           |
-|:------------|:-------------------|:------------------------------------------------------|
-| id          | String             | The Asset's unique identifier.                        |
-| accountId   | String             | The Asset's owning Centrapay Account id.              |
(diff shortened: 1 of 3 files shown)


src/api/assets/assets.md
```diff
@@ -66,9 +67,6 @@ Money assets have the following fields along with the base asset fields.
 | Field    | Type               | Description                                                          |
 |:---------|:-------------------|:---------------------------------------------------------------------|
 | currency | String             | Currency code, eg "NZD"                                              |
-| balance  | {% dt BigNumber %} | Current balance in the currency's smallest denomination (ie. cents). |
```
heymatthew, line 69: Can we keep this field around?

Because we don't have EXPERIMENTAL on it, we need to keep the API stable :)

heymatthew (approved):
LGTM

PR #346, opened 2022-03-03
Update Invitation docs

MeganSteenkamp:
* Add accepted and updated fields
* Change invitationCode to code
* Add `accountId` as a param required for accepting an Invitation
* Fix CRNs to be more realistic

### DIFF
--- a/src/api/invitations.md
+++ b/src/api/invitations.md
@@ -31,16 +31,21 @@ An Invitation can be used to allow users to claim ownership of a resource on the
 | expiresAt    | {% dt Timestamp %} | When the Invitation expires.                                             |
 | createdAt    | {% dt Timestamp %} | When the Invitation was created.                                         |
 | createdBy    | {% dt CRN %}       | The User or API Key that created the Invitation.                         |
+| accepted     | Boolean            | A flag indicating whether the Invitation has been accepted.              |
+| acceptedAt   | {% dt Timestamp %} | When the Invitation was accepted.                                        |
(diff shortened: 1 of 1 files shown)


src/api/invitations.md
```diff
@@ -54,18 +59,22 @@ An Invitation can be used to allow users to claim ownership of a resource on the
...
   "createdAt": "2021-08-25T00:02:49.488Z",
-  "createdBy": "crn::user:1234"
+  "createdBy": "crn:WIj211vFs9cNACwBb04vQw:api-key:MyApiKey",
+  "updatedAt": "2021-08-25T00:02:49.488Z",
+  "updatedBy": "crn:WIj211vFs9cNACwBb04vQw:api-key:MyApiKey"
 }
 {% endjson %}
 
 ### Accept an Invitation **EXPERIMENTAL**
+The Account id of the user that is accepting the Invitation must be supplied.
```
MeganSteenkamp, line 69: It seems like we should have an explanation here as accountId is not on the model.

IsaacMurtagh: We should be adding fields as a table with the name, type and description. It's not the same as the model of the resource. See https://docs.centrapay.com/api/accounts#create-an-account for an example of this. can we do this here?

MeganSteenkamp: @IsaacMurtagh have updated it, thanks

IsaacMurtagh (approved):

PR #359, opened 2022-03-21
Batches json nl

heymatthew:
Updated batch type from `farmlands` to `farmlands-external-asset` to make it more specific. We've got a generic endpoint so need to get specific on the types. Files have also been moved and redirects put in place.

If we don't do this, then we'll have a hard time distinguishing batches form the names.

Screen Shot 2022-03-22 at 5 04 47 PM

Updated table to use batch types. We need something that matches our documentation, before this was just a set of descriptions and it was hard to know what context was what.

Screen Shot 2022-03-22 at 5 05 38 PM

Made reference to JSONL format in top of the file

Screen Shot 2022-03-22 at 5 06 10 PM

...and created an inline example to make it easier to recognize. Models are still multi-line, there's a callout to make it clear that the example is laid out for legibility.

Screen Shot 2022-03-22 at 5 06 49 PM

### DIFF
--- a/src/api/batch/farmlands.md
+++ b/src/api/batch/farmlands.md
@@ -17,7 +17,7 @@ Loads Farmlands Card data into Centrapay as external assets.
 
 | Type Name   | farmlands-external-asset |
 | File Format | [JSONL][]{:.external}    |
-| File Schema | Array of [Account]       |
+| Record Type | [Account]       |
 
 ## Contents
 {:.no_toc .text-delta}
--- a/src/api/batch/farmlands.md
+++ b/src/api/batch/farmlands.md
@@ -5,17 +5,19 @@ parent: Batches
 grand_parent: Api Reference
 has_children: false
 nav_exclude: true
-permalink: /api/batch-types/farmlands
 type: farmlands-external-asset
+permalink: /api/batch-types/farmlands-external-asset
+redirect_from:
+  - /api/batch-types/farmlands
 ---
 
 # Farmlands External Asset Batch
 
 Loads Farmlands Card data into Centrapay as external assets.
 
-| Type Name   | farmlands          |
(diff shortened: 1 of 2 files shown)


src/api/batch/farmlands.md
```diff
@@ -164,159 +183,58 @@ with it. Note, this renders the availableBalance un-spendable.
...
-            status: "inactive",
-            expiry: "2022-08-01T22:32:56.631Z",
-          }
-        ]
-      }
-    ]
-  }
-]
-{% endjson %}
-
```
heymatthew, line 317: I moved the full example above so people can't miss the real format. This had to be JSONL to make sense in the context of this page.

src/api/batch/farmlands.md
```diff
@@ -89,69 +89,88 @@ Exported Farmlands Credit Card information used for importing and updating of a
...
+```
 
 <a name="example">
-## Examples
+## Example Models
+
+Centrapay matches perviously created resources using `externalId` and records
```
ncjones, line 108:
```suggestion
Centrapay matches previously created resources using `externalId` and records
```

src/api/batch/farmlands.md
```diff
@@ -5,18 +5,19 @@ parent: Batches
...
 # Farmlands External Asset Batch
 
 Loads Farmlands Card data into Centrapay as external assets.
 
-| Type Name   | farmlands          |
-| File Format | JSON               |
-| File Schema | Array of [Account] |
+| Type Name   | farmlands-external-asset |
+| File Format | [JSONL][]{:.external}    |
+| File Schema | Array of [Account]       |
```
ncjones, line 20:
```suggestion
| Record Type | [Account]       |
```

ncjones (approved):

PR #386, opened 2022-04-12
Update available balance to indicate it will default to 0 if not prov…

daallen303:
…ided

### DIFF
--- a/src/api/batch/farmlands-external-asset.md
+++ b/src/api/batch/farmlands-external-asset.md
@@ -38,7 +38,7 @@ Exported Farmlands Account used for importing and updating of a [Centrapay Accou
 | :--------------- | :--------------------------- | :-------------------------------------------------------- |
 | externalId       | String                       | Id used for keeping imported Centrapay Account up to date |
 | accountNumber    | String                       | Farmlands identifier attached to the Centrapay Account    |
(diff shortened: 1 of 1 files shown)


src/api/batch/farmlands-external-asset.md
```diff
@@ -38,7 +38,7 @@ Exported Farmlands Account used for importing and updating of a [Centrapay Accou
 | :--------------- | :--------------------------- | :-------------------------------------------------------- |
 | externalId       | String                       | Id used for keeping imported Centrapay Account up to date |
 | accountNumber    | String                       | Farmlands identifier attached to the Centrapay Account    |
-| availableBalance | {% dt BigNumber %} {% opt %} | Current spendable balance of Farmlands Ledger in cents                                    |
+| availableBalance | {% dt BigNumber %} {% opt %} | Current spendable balance of Farmlands Ledger in cents (defaults to 0)                    |
```
heymatthew, line 41: This doesn't match the way we're presenting in the rest of the table. Could we use a tone that matches contracts?

```suggestion
| availableBalance | {% dt BigNumber %} {% opt %} | Current spendable balance of Farmlands Ledger in cents. When not specified, defaults to 0.                    |
```

heymatthew (approved):
LGTM, just consistency and we're good.

PR #394, opened 2022-04-14
removed unused error code on create payment request

AndyClifford:


ncjones:
This is in use.

ncjones (approved):

ncjones:
I see this is in use for pay but not create.

PR #395, opened 2022-04-18
NO_AVAILABLE_PAYMENT_OPTIONS can now be thrown from bad barcode input

AndyClifford:


ncjones:
Should say something like "no payment options match the requested payment parameters". Patron code is only one of the inputs that can result in all payment options being filtered out. We will also filter out options on preauth=true and conditions=false.

tom-blake (approved):

PR #397, opened 2022-04-20
Add asset id to bank accounts and fix list bank accounts response

IsaacMurtagh:


src/api/bank-accounts/bank-accounts.md
```diff
@@ -60,6 +60,7 @@ bank transaction can be used to verify a bank account.
 | fullName     | String  | The first and last name of the user.                        |
 | emailAddress | String  | The user's email address.                                   |
 | test         | Boolean | A flag which is present if the Bank Account is for testing. |
+| assetId      | String  | The id of an asset which is associated to the bank account. |
```
MeganSteenkamp, line 63: This could have `Asset` with capitals as it is describing an Asset and then link to the Asset docs perhaps?

MeganSteenkamp:
Based on the title I was expecting to see changes to the list assets response?

IsaacMurtagh:
> Based on the title I was expecting to see changes to the list assets response?

ncjones (approved):

PR #411, opened 2022-05-03
Add barcode hint to payment request

alexBinns:


src/api/payment-requests/payment-requests.md
```diff
@@ -301,13 +302,14 @@ Payment Activities are created when a Payment Request has been **created**, **pa
...
 | :------------------- | :---------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | configId             | String            | The [Merchant Config][] id used to configure the payment options.                                                                                                        |
 | value                | {% dt Monetary %} | The canonical value of the payment request. Must be positive.                                                                                                            |
 | barcode              | String {% opt %}  | The patron's barcode to identify the account to attach the payment request to                                                                                            |
+| barcodeHint          | String {% opt %}  | **EXPERIMENTAL** Type of barcode provided. Required when barcode is present. Values are "farmlands" or "centrapay".                                                      |
```
IsaacMurtagh, line 312: Making this required when barcode is present will break the API contract.
And does paypal pass through a barcode when making a payment request, how will this effect them.

src/api/payment-requests/payment-requests.md
```diff
@@ -251,7 +251,8 @@ Payment Activities are created when a Payment Request has been **created**, **pa
   example {
     title 'Create a Payment Request with a Patron Code'
     body ({
-      barcode: '1219210961929460',
+      barcode: '458028560',
+      barcodeHint: 'farmlands',
```
IsaacMurtagh, line 255: The title of this one is with a patron code, which is described as a 16 digit number in our docs. I think we should keep this example using our Centrapay patron codes and not farmlands barcodes.

PR #427, opened 2022-05-24
Add accept payment condition as an operation on payment requests

IsaacMurtagh:


src/api/payment-requests/payment-requests.md
```diff
@@ -829,6 +835,41 @@ descending activity created date.
...
 {% endjson %}
 
+<a name="accept-payment-condition"></a>
+### Accept Payment Condition For a Payment Request **EXPERIMENTAL**
+
+Accept one of the payment condition listed in the `merchantConditions` field of a Payment Request
+that has status `awaiting-merchant`. Returns a [Payment Activity][] in the response.
+
+{% reqspec %}
+  POST '/api/payment-requests/{paymentRequestId}/condition/{conditionId}/accept'
```
MeganSteenkamp, line 845: Were you going to update this to `conditions`? :)

src/api/payment-requests/payment-requests.md
```diff
@@ -104,20 +104,23 @@ version (documented on this page) and the "legacy" version (documented at
 ★  For payment options which specify an address, there's a requirement to make a transaction on an external ledger.
 Once you have made that payment, you can use the transaction id to [Pay a Payment Request](#pay) using the legacy payment API.
 
-<a name="merchant-condition">
-### Merchant Condition
-Some asset types may require conditional operator approval. Requires `conditionsEnabled` to be set to true or the asset type may not be a [Payment Option](#payment-option).
+<a name="payment-condition">
+### Payment Condition
```
MeganSteenkamp, line 108: I had thought about changing this to Payment Condition but then I think you need to change the description below that says `Some asset types may require conditional operator approval.` as this is specifically talking about merchant conditions.

src/api/payment-requests/payment-requests.md
```diff
@@ -829,6 +835,41 @@ descending activity created date.
 }
 {% endjson %}
 
+<a name="accept-payment-condition"></a>
+### Accept Payment Condition For a Payment Request **EXPERIMENTAL**
+
+Accept one of the payment condition listed in the `merchantConditions` field of a Payment Request
+that has status `awaiting-merchant`. Returns a [Payment Activity][] in the response.
```
MeganSteenkamp, line 842: If you are changing the above to Payment Condition this could just describe accepting a payment condition and not specifically reference merchant conditions.

IsaacMurtagh: I was thinking we just add patronConditions and awaiting-patron when they're valid fields as opposed to emitting the specifics. Otherwise, there is nothing saying when a condition can be accepted.

src/api/payment-requests/payment-requests.md
IsaacMurtagh: Yep I'll make it more general 🙂

src/api/payment-requests/payment-requests.md
```diff
@@ -829,6 +835,41 @@ descending activity created date.
...
+  createdBy: "crn::user:0af834c8-1110-11ec-9072-3e22fb52e878",
+  paymentRequestCreatedBy: "crn::user:0af834c8-1110-11ec-9072-3e22fb52e878",
+  activityNumber: "2",
+  merchantAccountId: "C4QnjXvj8At6SMsEN4LRi9",
+  merchantId: "5ee0c486308f590260d9a07f",
+  merchantConfigId: "5ee168e8597be5002af7b454",
+  merchantName: "Centrapay Café",
+}
+{% endjson %}
+
```
MeganSteenkamp, line 872: I think we are missing the error codes that can be thrown from this endpoint.

src/api/payment-requests/payment-requests.md
```diff
@@ -393,7 +399,7 @@ Payment Activities are created when a Payment Request has been **created**, **pa
   ],
   "merchantConditions": [
     {
-      "id": "1",
+      "id": 1,
```
MeganSteenkamp, line 402: This is returned as a string so I think how it was is correct.

MeganSteenkamp (approved):

PR #435, opened 2022-06-06
V2 cancel

laurencemcknight:
We require a non-legacy cancel endpoint for pre-auth payment requests.
This PR documents the new V2 endpoint.

### DIFF
--- a/src/api/payment-requests/payment-requests.md
+++ b/src/api/payment-requests/payment-requests.md
@@ -719,154 +719,76 @@ Alternatively you can provide an external transaction Id and the Centrapay [Asse
 {% reqspec %}
   POST '/api/payment-requests/{paymentRequestId}/cancel'
   auth 'jwt'
+  path_param 'paymentRequestId', 'MhocUmpxxmgdHjr7DgKoKw'
 {% endreqspec %}
 
 {% h4 Example response payload %}
 {% json %}
 {
-  "id": "MhocUmpxxmgdHjr7DgKoKw",
-  "shortCode": "CPC7FZS5015",
-  "url": "https://app.centrapay.com/pay/MhocUmpxxmgdHjr7DgKoKw",
-  "patronCodeId": "V17FByEP9gm1shSG6a1Zzx",
-  "barcode": "9990001234567895",
-  "merchantId": "26d3Cp3rJmbMHnuNJmks2N",
+  "type": "cancelled",
+	"cancellationReason": "CANCELLED_BY_MERCHANT",
+  "value": {
+    "currency": "NZD",
+    "amount": "1000",
+  },
+  "assetType": "centrapay.nzd.main",
+  "paymentRequestId": "MhocUmpxxmgdHjr7DgKoKw",
   "merchantName": "Centrapay Café",
-  "configId": "5efbe2fb96c08357bb2b9242",
-  "purchaseOrderRef": "oF6kj1QlH5gK0y9rjRHFh2",
-  "invoiceRef": "sy8CRmo3sp3ArOpnfmb423",
-  "value": { "currency": "NZD", "amount": "6190" },
-  "paymentOptions": [
-    {
-      "amount": "6190",
-      "assetType": "centrapay.nzd.test"
-    },
-    {
-      "amount": "6190",
-      "assetType": "epay.nzd.test",
-      "productCodes": [ "23403" ]
-    }
-  ],
-  "lineItems": [
-      {
-        "name": "Coffee Grounds",
-        "sku": "GH1234",
-        "qty": "1",
-        "price": "4195",
-        "tax": "15.00",
-      },
-      {
-        "name": "Centrapay Cafe Mug",
-        "sku": "SB456",
-        "qty": "25",
-        "price": "1995",
-        "tax": "15.00",
-        "discount": "199",
-      },
-  ],
-  "merchantConditions": [
-    {
-      "id": "1",
-      "name": "photo-id-check",
-      "message": "Please check ID",
-      "status": "awaiting-merchant"
-    }
-  ],
-  "status": "cancelled",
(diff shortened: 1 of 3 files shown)


src/api/payment-requests/payment-requests.md
```diff
@@ -713,6 +713,83 @@ Alternatively you can provide an external transaction Id and the Centrapay [Asse
...
+{% h4 Error Responses %}
+
+| Status |             Code              |                             Description                              |
+| :----- | :---------------------------- | :------------------------------------------------------------------- |
+| 403    | {% break _ REQUEST_EXPIRED %} | The payment request has expired and therefore cannot be cancelled.   |
+| 403    | {% break _ REQUEST_PAID %}    | The payment request has been paid and therefore cannot be cancelled. |
+
+<a name="void">
+### Void a Payment Request **EXPERIMENTAL**
+
```
laurencemcknight, line 756: Do we need to explain that void chooses between cancelling and refunding a payment request?

src/api/payment-requests/payment-requests.md
```diff
@@ -713,6 +713,83 @@ Alternatively you can provide an external transaction Id and the Centrapay [Asse
...
+{% reqspec %}
+  POST '/api/payment-requests/{paymentRequestId}/void'
+  auth 'jwt'
+  path_param 'paymentRequestId', 'MhocUmpxxmgdHjr7DgKoKw'
+{% endreqspec %}
+
+{% h4 Example response payload %}
+{% json %}
+{
+  "type": "voidCancel",
```
laurencemcknight, line 766: The legacy endpoint returned types `voidCancel` and `voidRefund`. For this new endpoint could we instead return the `cancelled` and `refunded` types that are already used by the V2 endpoints?

src/api/payment-requests/payment-requests.md
```diff
@@ -713,6 +713,83 @@ Alternatively you can provide an external transaction Id and the Centrapay [Asse
...
+  "paymentRequestCreatedBy": "crn::user:0af834c8-1110-11ec-9072-3e22fb52e878",
+  "activityNumber": "2",
+}
+{% endjson %}
+
+{% h4 Error Responses %}
+
+| Status |                Code                |                                           Description                                            |
+| :----- | :--------------------------------- | :----------------------------------------------------------------------------------------------- |
+| 403    | {% break _ VOID_WINDOW_EXCEEDED %} | The time since the payment request was created exceeds the time a payment request can be voided. |
```
laurencemcknight, line 789: Is it worthwhile getting Jan's thoughts on void windows, or are we safe to use the existing 24 hour window?

src/api/payment-requests/payment-requests.md
heymatthew: It might be better to use language consistent with what we have in the [guides section](https://docs.centrapay.com/guides/merchant-terminals) of the docs :).

Details aren't as important. This is an example of rest without put so think of it as a business process that we're applying to the payment request.

heymatthew: Can we break out void and cancel as separate PRs?

Cancel is easier to talk through :)

src/api/payment-requests/payment-requests.md
heymatthew: Lets separate this into another PR.

Lets raise an ADR to
* Make void a status returned on the new APIs
* Derive voidCancel and voidRefund based on activities of the loaded in PaymentRequest
* Save into Dynamo the payment request as we want it to be moving forward.

src/guides/merchant-terminals.md
```diff
@@ -140,7 +140,7 @@ Request is paid.
 [Payment Requests]: {% link api/payment-requests/payment-requests.md %}
 [Merchant Configs]: {% link api/merchants/merchant-configs.md %}
 [Get Payment Request]: {% link api/payment-requests/legacy-payment-requests.md %}#requests-info
-[Void Payment Request]: {% link api/payment-requests/legacy-payment-requests.md %}#requests-void
+[Void Payment Request]: {% link api/payment-requests/payment-requests.md %}#void
```
heymatthew, line 143: you can just use `#void` here, we're not on a different page

src/api/payment-requests/payment-requests.md
```diff
@@ -713,6 +713,83 @@ Alternatively you can provide an external transaction Id and the Centrapay [Asse
...
+  "activityNumber": "2",
+}
+{% endjson %}
+
+{% h4 Error Responses %}
+
+| Status |             Code              |                             Description                              |
+| :----- | :---------------------------- | :------------------------------------------------------------------- |
+| 403    | {% break _ REQUEST_EXPIRED %} | The payment request has expired and therefore cannot be cancelled.   |
+| 403    | {% break _ REQUEST_PAID %}    | The payment request has been paid and therefore cannot be cancelled. |
```
heymatthew, line 752: Have a read of this: https://sive.rs/book/OnWritingWell

expiresAt has passed based on the configured expirySeconds

src/api/payment-requests/payment-requests.md
```diff
@@ -713,6 +713,44 @@ Alternatively you can provide an external transaction Id and the Centrapay [Asse
...
+  "activityNumber": "2",
+}
+{% endjson %}
+
+{% h4 Error Responses %}
+
+| Status |             Code              |                         Description                         |
+| :----- | :---------------------------- | :---------------------------------------------------------- |
+| 403    | {% break _ REQUEST_EXPIRED %} | expiresAt has passed based on the configured expirySeconds. |
+| 403    | {% break _ REQUEST_PAID %}    | status is `paid`.                                           |
```
heymatthew, line 752: Prefer sentence case, most tables don't use this short form.

heymatthew (approved):

PR #451, opened 2022-07-14
Fix broken links

heymatthew:
Came across broken links, fixed them where I could

We should lint for this, matching on `]()` should throw an error.


ae8698ae (approved):
LGTM

alexBinns (approved):

PR #465, opened 2022-08-05
Add preAuthStatus to Payment Request docs

daallen303:


src/api/payment-requests/payment-requests.md
```diff
@@ -586,6 +587,7 @@ Payment Activities are created when a Payment Request has been **created**, **pa
   "liveness": "main",
   "expirySeconds": 120,
   "preAuth": true,
+  "preAuthStatus": "expired",
```
laurencemcknight, line 590: We are not going to use 'expired' as a preAuthStatus, we will use 'released' instead.

laurencemcknight (changes requested):

src/api/payment-requests/payment-requests.md
daallen303: Yep, this has been changed.

laurencemcknight (approved):

PR #475, opened 2022-08-12
Add documentation for void confirmation endpoint

laurencemcknight:
This adds documentation for the upcoming /void-confirmation endpoint.

### DIFF
--- a/src/api/payment-requests/payment-requests.md
+++ b/src/api/payment-requests/payment-requests.md
@@ -1109,11 +1109,11 @@ Voiding a confirmation will cancel the confirmation and return any transferred f
 
 {% h4 Error Responses %}
 
-| Status |                Code                |                               Description                                |
-| :----- | :--------------------------------- | :----------------------------------------------------------------------- |
-| 403    | {% break _ VOID_WINDOW_EXCEEDED %} | The void window is closed 24 hours after the confirmation's `createdAt`. |
-| 403    | {% break _ PRE_AUTH_EXPIRED %}     | `preAuthExpiresAt` has passed                                            |
-| 403    | {% break _ REQUEST_CANCELLED %}    | The Payment Request has been cancelled                                   |
+| Status |                Code                |                        Description                        |
+| :----- | :--------------------------------- | :-------------------------------------------------------- |
+| 403    | {% break _ VOID_WINDOW_EXCEEDED %} | 24 hours have passed since the confirmation's `createdAt` |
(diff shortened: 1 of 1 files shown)


src/api/payment-requests/payment-requests.md
```diff
@@ -248,15 +248,16 @@ Payment Activities are created when a Payment Request has been **created**, **pa
...
+| request           | [Payment Request][] was created.                                                                              |
+| payment           | [Payment Request][] was paid.                                                                                 |
+| refund            | Funds were returned to the shopper.                                                                           |
+| cancellation      | [Payment Request][] was cancelled by the merchant or the shopper.                                             |
+| expiry            | [Payment Request][] wasn't paid before time out.                                                              |
+| accept-condition  | A [Payment Condition][] was accepted.                                                                         |
+| decline-condition | A [Payment Condition][] was declined.                                                                         |
+| void-confirmation | A Confirmation Activity was cancelled by the merchant and any transferred funds were returned to the shopper. |
```
heymatthew, line 260: AFAIK funds are not returned to the shopper. This is under active discussion, but here's a go at some other words:

```suggestion
| void-confirmation | A Confirmation was reversed by the merchant. |
```

heymatthew (changes requested):
Can you update this language across this document please? Confirmation and Void Confirmation don't move funds.

src/api/payment-requests/payment-requests.md
```diff
@@ -1070,6 +1071,50 @@ Confirmations require an `idempotencyKey` in order to ensure that a confirmation
 | 403    | {% break _ INVALID_AMOUNT   %}             | The confirmation is greater then the remaining funds on the authroization                                                                                           |
 | 403    | {% break _ IDEMPOTENT_OPERATION_FAILED %}  | There has already been a confirmation against the Payment Request with the same idempotencyKey but different content. We recommend that you void this confirmation. |
 
+<a name="void-confirmation"></a>
+### Void a confirmation against a Pre Auth Payment Request **EXPERIMENTAL**
+
+Voiding a confirmation will reverse the confirmation. An `idempotencyKey` is required as a reference to the confirmation to be voided.
```
heymatthew, line 1077: We should document a 200 here even when we don't know what the idempotency key means. So if a void-confirm comes through, we'd say something like "We've done everything we can with this in our system to back out any confirmation associated with that idempotency key".

heymatthew:
It looks like the designs are taking us in another direction. Closing this PR.

PR #479, opened 2022-08-22
Nuxt3 Hello World

AndyClifford:
Created a Hello World Nuxt3 project to be deployed alongside the existing docs at /v2

**Test Plan**: go to docs.centrapay.com/v2 and check that Hello World is delivered and the button routes to docs.centrapay.com


build-nuxt-site.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
```
ncjones, line 2: Every bash script should start with `set -euo pipefail` so that errors are propagated the way you expect they should. In this case probably this script is failing because the nodejs dependencies have not been installed.

.github/workflows/publish.yml
```diff
@@ -23,6 +23,7 @@ jobs:
       - run: bundle install
       - run: bundle exec jekyll build
       - run: cp CNAME _site
+      - run: ./build-nuxt-site.sh
```
ncjones, line 26: This needs to be added to both workflow files: 

.github/workflows/publish.yml
.github/workflows/validate.yml

ncjones: I think you need to add:

```
      - uses: actions/setup-node@v2
        with:
          node-version: '16'
      - run: yarn install --frozen-lock-file
```

new-docs/nuxt.config.ts
```diff
@@ -0,0 +1,6 @@
+import { defineNuxtConfig } from 'nuxt'
```
ncjones, line 1: This file should have `.js` or `.mjs` extension.

new-docs/pages/v2.vue
```diff
@@ -0,0 +1,22 @@
...
+    <h1>{{ name }}</h1>
+    <button @click="routeToDocs">Old Docs</button>
+  </div>
+</template>
+
+<script>
+export default {
+  data() {
+    return {
+      name: 'Hello World'
```
ncjones, line 12: Maybe in a follow-up PR, but this should be linted with our normal eslint settings.

new-docs/README.md
```diff
@@ -0,0 +1,42 @@
+# Nuxt 3 Minimal Starter
```
ncjones, line 1: Don't need this file.

build-nuxt-site.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
tom-blake, line 1: We need to improve local DX in a follow up 👍

new-docs/pages/v2.vue
AndyClifford: We have added a followup task to add linting 👍

tom-blake (approved):

PR #481, opened 2022-08-23
Add Linter to Nuxt3 Project

AndyClifford:
ESLint, plugins and existing config have been copied from Ariel. Added the nuxt eslint plugin as a dependency and included the recommended rules in the config file to begin with. Linting has also been added as a pipeline step so there are some other very minor changes to accompany this PR

### DIFF
--- a/build-nuxt-site.sh
+++ b/build-nuxt-site.sh
@@ -4,8 +4,6 @@ set -euo pipefail
 
 cd new-docs
 
-yarn install --frozen-lock-file
-
 yarn generate
 
 rm .output/public/index.html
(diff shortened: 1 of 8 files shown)


build-nuxt-site.sh
```diff
@@ -6,6 +6,8 @@ cd new-docs
 
 yarn install --frozen-lock-file
 
+yarn lint
```
Ray2126, line 9: Should this be a seperate step? Linting isn't part of building the nuxt site.

Ray2126 (approved):

PR #487, opened 2022-08-23
Add void error for pre auth with confirmations

alexBinns:
Ths PR adds a new error to void for pre auth payment requests that have confirmations loaded against them

### DIFF
--- a/src/api/payment-requests/payment-requests.md
+++ b/src/api/payment-requests/payment-requests.md
@@ -904,14 +904,15 @@ Voiding a payment request will cancel the request and trigger any refunds if nec
 
 {% h4 Error Responses %}
 
-| Status |                Code                |                                                                                                                     Description                                                                                                                     |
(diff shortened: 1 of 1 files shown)


src/api/payment-requests/payment-requests.md
```diff
@@ -904,14 +904,15 @@ Voiding a payment request will cancel the request and trigger any refunds if nec
...
+| 403    | {% break _ REFUND_NOT_SUPPORTED %}       | The asset type does not support refunds.                                                                                                                                                                                                            |
+| 403    | {% break _ REQUEST_EXPIRED %}            | The payment request has expired.                                                                                                                                                                                                                    |
+| 403    | {% break _ PRE_AUTH_ALREADY_CONFIRMED %} | The pre auth payment request already has confirmations. We recommend using refund to reverse the transaction.                                                                                                                                                        |
```
heymatthew, line 915: Lets get rid of "We recommend" and use active tone.

Lets get rid of the "Void this confirmation" from elsewhere in the document.

Can we use capitals on the nouns here?

heymatthew (approved):

PR #489, opened 2022-08-24
Create home page for new docs site

Ray2126:
Test plan:

- Go to docs.centrapay.com/v2 and see that it looks like 
image


tom-blake (approved):

AndyClifford (approved):

new-docs/components/Card.vue
```diff
@@ -0,0 +1,43 @@
+<template>
+  <div
```
MeganSteenkamp, line 2: @Ray2126 I know this is merged but some feedback would be that this should be an anchor tag 👍

AndyClifford: Would this avoid needing an onClick method?

MeganSteenkamp: @AndyClifford yes

Ray2126: Thanks @MeganSteenkamp , I have a PR up to replace the div with a `NuxtLink` which is Nuxt's version of an anchor tag.

Will change the `h6` and `header` tags as well :)

PR #493, opened 2022-08-29
Description and examples for the name of a Classification

laurencemcknight:
The name property on the classification model had a description that a referenced a non existent 'description' property on the GS1 global product classification list. This description has been updated to instead reference the 'title' property in GS1.

The name property on the classification model is a required field but was not included in any of the curl examples. The examples have been updated to include the name property.

### DIFF
--- a/src/api/payment-requests/payment-requests.md
+++ b/src/api/payment-requests/payment-requests.md
@@ -171,7 +171,7 @@ line items may be represented as a separate line item with a negative amount.
 | ----- | ------ | ------------------------------------ |
 | type  | String | The classification type (see below). |
 | code  | String | The classification code.             |
-| name  | String | The classification description.      |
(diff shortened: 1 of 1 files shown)


src/api/payment-requests/payment-requests.md
```diff
@@ -354,6 +354,7 @@ Payment Activities are created when a Payment Request has been **created**, **pa
           classification: {
             type: 'GS1',
             code: '10001874',
+            name: 'CROCKERY',
```
laurencemcknight, line 357: I chose this as the example as in the GS1 examples:
1. Titles are capitalised 
2. Titles describe the category of a product

heymatthew (approved):

heymatthew (approved):

PR #496, opened 2022-09-05
Move remaining ruby files into the legacy directory

Ray2126:
**Test plan:**
- Regression test: [docs.centrapay.com](url) should still show the new landing page. [docs.centrapay.com/guides](url) and [docs.centrapay.com/api-reference](url) should not have changed.


docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
     image: ruby:2.7
     working_dir: /work
     volumes: [ ".:/work" ]
-    command: "bundle exec jekyll serve --livereload --host=0.0.0.0"
+    command: "bash -c 'cd legacy && bundle exec jekyll serve --livereload --host=0.0.0.0'"
```
ncjones, line 8: just change the working_dir

.github/workflows/publish.yml
```diff
@@ -13,15 +13,16 @@ jobs:
...
-          path: vendor/bundle
+          path: legacy/vendor/bundle
           key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
           restore-keys: |
             ${{ runner.os }}-gems-
       - run: gem install bundler
-      - run: bundle install
-      - run: bundle exec jekyll build
+      - run: cd legacy && bundle install
+      - run: cd legacy && bundle exec jekyll build
```
ncjones, line 25: If you wrap the yarn and jekyll builds in a build.sh then the build command is the same for Github, local, and Netlify. In the netlify toml you can define a custom build command and set the ruby and node versions.

AndyClifford (approved):

PR #515, opened 2022-09-15
Add container to serve the built Nuxt site

AndyClifford:
This container serves the same assets as the production site. We have found there are some differences between the development Nuxt app and the built production site. This will allow better local testing.


package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "scripts": {
     "lint": "vue-cli-service lint --max-warnings 0 --no-fix .",
-    "start": "ENV=development nuxt dev --no-clear",
+    "start": "nuxt generate && ENV=development nuxt dev --no-clear",
```
Ray2126, line 5: You need the `ENV` variable for `nuxt generate` as well

Ray2126 (approved):

PR #528, opened 2022-10-10
Add mobile support for table of contents

tom-blake:
Our table of contents in our guides template now appears as a sticky dropdown on smaller screens (see screenshot).

Nuxt Content's document driven mode has been turned on to take advantage of a built-in table of contents composable that provides a ready-to-use data structure containing our markdown headings for us. There are also prev/next and navigation components that we can use in the future.

More on document-driven mode here https://content.nuxtjs.org/guide/writing/document-driven/

Test plan:
- Confirm table of contents switches to drop down menu at screen size
  drops below large breakpoint
- Confirm table of content behavior has not changed for >large screen
  sizes
- Confirm headings are visible when scrolling finishes after clicking a
  table of contents link when screen size is less than mobile breakpoint

### Default ToC view
image

### Expanded ToC
image

### Headings are visible after clicking ToC links 
image

### DIFF
--- a/components/Layout.vue
+++ b/components/Layout.vue
@@ -1,18 +1,16 @@
 <template>
-  <div class="relative mx-auto flex max-w-8xl justify-center sm:px-2 lg:px-8 xl:px-12">
-    <div class="hidden lg:relative lg:block lg:flex-none">
-      <div class="absolute inset-y-0 right-0 w-[50vw] bg-slate-50 dark:hidden" />
-      <div class="sticky top-[4.5rem] -ml-0.5 h-[calc(100vh-4.5rem)] overflow-y-auto py-16 pl-0.5">
-        <div class="absolute top-16 bottom-0 right-0 hidden h-12 w-px bg-gradient-to-t from-slate-800 dark:block " />
-        <div class="absolute top-28 bottom-0 right-0 hidden w-px bg-slate-800 dark:block" />
+  <div class="mx-auto max-w-7xl flex flex-col-reverse min-h-full lg:grid lg:grid-cols-5 lg:gap-8 lg:px-8">
+    <div class="hidden lg:relative lg:block lg:flex-none lg:col-span-1">
+      <div class="absolute inset-y-0 right-0 w-[50vw] bg-slate-50" />
+      <div class="sticky top-[4.5rem] h-[calc(100vh-4.5rem)] overflow-y-auto py-16 px-4">
         <Navigation
           :navigation="navigation"
-          class="w-64 pr-8 xl:w-72 xl:pr-16"
+          class="w-full"
         />
       </div>
     </div>
     <div
-      class="min-w-0 max-w-2xl flex-auto px-4 py-16 lg:max-w-none lg:pr-0 lg:pl-8 xl:px-16"
+      class="min-w-full max-w-2xl flex-auto px-4 py-3 lg:py-16 lg:max-w-prose lg:col-span-3"
     >
       <article>
         <header
@@ -27,7 +25,7 @@
           </p>
           <h1
             v-if="title"
-            class="font-display font-bold text-4xl tracking-tight text-slate-900 dark:text-white"
+            class="font-display font-bold text-4xl text-slate-900"
           >
             {{ title }}
           </h1>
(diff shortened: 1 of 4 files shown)


components/Layout.vue
```diff
@@ -1,52 +1,50 @@
...
           <dt
-            class="font-display text-sm font-medium text-slate-900 dark:text-white"
+            class="font-display text-sm font-medium text-slate-900"
           >
             Previous
           </dt>
           <dd class="mt-1">
             <NuxtLink
               :href="previousPage._path"
-              class="text-base font-semibold text-slate-500 hover:text-slate-600 dark:text-slate-400 dark:hover:text-slate-300"
```
tom-blake, line 49: I've removed `dark` modifiers since we don't support dark mode yet

pages/[...slug].vue
```diff
@@ -1,18 +1,7 @@
 <template>
   <main>
-    <Layout
-      :title="title"
-      :table-of-contents="tableOfContents"
```
Comment (tom-blake) [inline pages/[...slug].vue:5]: I've localised these props inside the Layout component because we weren't using them anywhere else

components/Layout.vue
```diff
@@ -142,6 +148,8 @@ const nextPage = allLinks[linkIndex + 1];
 const section = navigation.find((s) =>
   s.children.find((child) => child._path === path)
 );
+const { toc } = useContent();
```
tom-blake, line 151: Here's where I'm using the [document driven composable](https://content.nuxtjs.org/api/composables/use-document-driven)

components/Layout.vue
```diff
@@ -1,52 +1,50 @@
 <template>
-  <div class="relative mx-auto flex max-w-8xl justify-center sm:px-2 lg:px-8 xl:px-12">
-    <div class="hidden lg:relative lg:block lg:flex-none">
-      <div class="absolute inset-y-0 right-0 w-[50vw] bg-slate-50 dark:hidden" />
-      <div class="sticky top-[4.5rem] -ml-0.5 h-[calc(100vh-4.5rem)] overflow-y-auto py-16 pl-0.5">
-        <div class="absolute top-16 bottom-0 right-0 hidden h-12 w-px bg-gradient-to-t from-slate-800 dark:block " />
-        <div class="absolute top-28 bottom-0 right-0 hidden w-px bg-slate-800 dark:block" />
+  <div class="mx-auto max-w-7xl flex flex-col-reverse min-h-full lg:grid lg:grid-cols-5 lg:gap-8 lg:px-8">
```
tom-blake, line 2: Display behaviour:
+ Screen < large - `flex-col-reverse` puts the ToC at the top of the page
+ Screen >= large: Using a flex grid to size the 3 "columns" proportionally

components/Layout.vue
```diff
@@ -1,52 +1,50 @@
...
-      class="min-w-0 max-w-2xl flex-auto px-4 py-16 lg:max-w-none lg:pr-0 lg:pl-8 xl:px-16"
+      class="min-w-full max-w-2xl flex-auto px-4 py-3 lg:py-16 lg:max-w-prose lg:col-span-3"
     >
       <article>
         <header
           :v-if="title || section"
           class="mb-9 space-y-1"
         >
-          <p
+          <h5
```
Ray2126, line 20: An `h5` tag directly before an `h1` tag isn't great for accessibility. I don't know what the correct tag is here, maybe `span`?

components/Layout.vue
```diff
@@ -58,82 +56,90 @@
...
           </dd>
         </div>
       </dl>
     </div>
-    <div class="hidden xl:sticky xl:top-[4.5rem] xl:-mr-6 xl:block xl:h-[calc(100vh-4.5rem)] xl:flex-none xl:overflow-y-auto xl:py-16 xl:pr-6">
-      <nav
-        aria-labelledby="on-this-page-title"
-        class="w-56"
+    <nav
+      class="sticky top-[4.5rem] flex flex-col w-full px-4 bg-white lg:max-h-page lg:self-start lg:bg-transparent lg:backdrop-blur-none lg:flex-none lg:overflow-y-auto lg:py-16 lg:col-span-1"
```
Ray2126, line 75: Minor - a shadow here might be nice to give the TOC dropdown some depth. It's hard to tell where the dropdown ends and the content starts.

image

Ray2126 (approved):

components/Layout.vue
tom-blake: Good spot! I've reverted to a `p` tag for now - that should be ok

components/Layout.vue
tom-blake: I opted against this as it doesn't look great next to the drop shadow being applied to the header
image

tom-blake: A small drop shadow looks better - perhaps this is ok enough to leave in
image

Ray2126: Yep, I like the small drop shadow 😄

PR #529, opened 2022-10-13
Refactor previous/next navigation to use document driven composable

tom-blake:
Now that we've turned on document driven mode, we can take advantage of further abstractions available through Nuxt Content's built-in composables.

More on document-driven mode here: https://content.nuxtjs.org/guide/writing/document-driven More on useContent here: https://content.nuxtjs.org/api/composables/use-document-driven

Test plan:
- Confirm that previous/next page navigation redirects to the correct page

### DIFF
--- a/components/Layout.vue
+++ b/components/Layout.vue
@@ -35,7 +35,7 @@
       <dl
         class="mt-12 flex border-t border-slate-200 pt-6"
       >
-        <div v-if="previousPage">
+        <div v-if="prev">
           <dt
             class="font-display text-sm font-medium text-slate-900"
           >
@@ -43,16 +43,16 @@
           </dt>
           <dd class="mt-1">
             <NuxtLink
-              :href="previousPage._path"
+              :to="prev._path"
               class="text-base font-semibold text-slate-500 hover:text-slate-600"
             >
-              <span aria-hidden="true">&larr;</span> {{ previousPage.title }}
+              <span aria-hidden="true">&larr;</span> {{ prev.title }}
             </NuxtLink>
           </dd>
         </div>
 
         <div
-          v-if="nextPage"
+          v-if="next"
           class="ml-auto text-right"
         >
           <dt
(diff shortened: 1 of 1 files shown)


components/Layout.vue
```diff
@@ -139,14 +139,10 @@ const contentPath = path.endsWith('/') ? path.slice(0, -1) : path;
 const contentDirectory = await queryContent().where({ _path: contentPath }).findOne();
 const title = contentDirectory.title;
 const navigation = await fetchContentNavigation();
-const allLinks = navigation.flatMap((section) => section.children);
-const linkIndex = allLinks.findIndex((link) => link._path === path);
-const previousPage = allLinks[linkIndex - 1];
-const nextPage = allLinks[linkIndex + 1];
```
Ray2126, line 145: So satisfying removing all this logic! Nice one 💪

components/Layout.vue
```diff
@@ -139,14 +139,10 @@ const contentPath = path.endsWith('/') ? path.slice(0, -1) : path;
...
 const section = navigation.find((s) =>
   s.children.find((child) => child._path === path)
 );
-const { toc } = useContent();
+const { toc, prev: previousPage, next: nextPage } = useContent();
```
Ray2126, line 145: Any reason why we needed to rename the variables here?

tom-blake: These were the names already being used in the Vue template - that's why this diff is small 👍 

I like the idea of [including the section heading titles](https://content.nuxtjs.org/api/composables/use-document-driven) in the prev/next navigation too:

image

 I imagined these variables as `prevSectionTitle` (Component) and `nextSectionTitle` (Utils) respectively so more descriptive variables felt better here

tom-blake: @Ray2126 I've been playing around and I don't think #3 needs to be considered anymore - I've updated the variables to use the names provided by the Nuxt composable.

Please note the new additions to the diff

Ray2126 (approved):

PR #537, opened 2022-11-25
Add payment activities in the payment request response

daallen303:


daallen303:
We are no longer considering this change

PR #546, opened 2022-12-20
💅 Render mermaid diagrams using Markdown

MeganSteenkamp:
**🧜‍♀️ Render mermaid diagrams using Markdown 🧜‍♀️**
- This has basic styling and a diagram from an existing Notion guide as a POC.
- Multiple diagrams are able to be rendered on the same page.
- There is a layout shift when the diagram is rendered, this won't be able to be fixed with our current pattern as once the DOM tree is created we insert a node with the diagram.

- Design: https://www.figma.com/file/AuIWDdnDKEMXhxEJIU6UNP/Diagrams?node-id=346%3A2365&t=q6RL8chixiMoStp9-4

- Story: https://www.notion.so/centrapay/65a900a58a64424e85e3ccbaad674f51?v=f3704058540e48bc8c3680e61c76b2e3&p=07dbf6df16f343b29122be1cb4f859f3&pm=s

**Test plan**: 
- Expect to see mermaid diagram rendered in the v2 docs as per screenshot 🧜‍♀️
- Expect regular code blocks to work (regression test)

image

### DIFF
--- a/components/content/ProseCode.vue
+++ b/components/content/ProseCode.vue
@@ -1,0 +1,68 @@
+<script setup>
+import mermaid from 'mermaid';
+
+const props = defineProps({
+  code: {
+    type: String,
+    default: ''
+  },
+  language: {
+    type: String,
+    default: ''
+  },
+  filename: {
+    type: String,
+    default: null
+  },
+  highlights: {
+    type: Array,
+    default: () => []
+  }
+});
+
+onMounted(() => {
+  if (props.language === 'mermaid' && props.code) {
+    const parentNode = document.getElementById('mermaid-graph');
+    const mermaidContainer = document.createElement('div');
+    mermaidContainer.innerHTML = props.code;
+    parentNode.replaceWith(mermaidContainer);
+    mermaid.initialize({
+      startOnLoad: true,
+      theme: 'base',
+      themeVariables: {
+        mainBkg: '#F8F8F8',
+        actorBkg: '#F8F8F8',
+        primaryColor: '#F8F8F8',
+        primaryTextColor: '#232323',
+        primaryBorderColor: '#D1D5DB',
+        secondaryColor: '#232323',
+        noteBkgColor: '#232323',
+        noteTextColor: '#F8F8F8',
+        labelBoxBkgColor: '#FFFFFF',
+      },
+      sequence: {
+        diagramMarginY: 50,
+      }
+    });
+    mermaid.init(undefined, mermaidContainer);
+  }
+});
+</script>
+
+<template>
+  <div
+    class="my-8 rounded-lg"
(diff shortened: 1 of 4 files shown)


components/content/ProseCode.vue
```diff
@@ -0,0 +1,68 @@
...
+    type: String,
+    default: null
+  },
+  highlights: {
+    type: Array,
+    default: () => []
+  }
+});
+
+onMounted(() => {
```
AndyClifford, line 23: Will this hook be called for static site generation?

MeganSteenkamp: I saw that in the Vue docs that this hook not called when using server side rendering.
Are our docs a static site? If so, this hook will be called. 

https://nuxtjs.org/docs/features/rendering-modes#server-side-rendered-sites-and-static-sites

I have looked at the static site at port `:5001` and I can see the diagram as expected.

assets/css/colors.js
```diff
@@ -1,4 +1,7 @@
 const colors = {
+  core: {
+    'off-white': '#F2F2F2',
```
MeganSteenkamp, line 3: Junjie is using the `core` collection of colours in Figma for the designs.
I will add the set of tokens in a separate PR.

Slack thread: https://centrapay.slack.com/archives/CTM2483GW/p1671564336450279 🧵

components/content/ProseCode.vue
AndyClifford: Sweet. I'm not really sure if ours is ssr or static to be honest. We don't have `ssr: false` or [target: static](https://nuxtjs.org/docs/features/deployment-targets/) anywhere in our codebase but we still use `yarn generate` which outputs a bunch of static files

image

AndyClifford: I remember our initial plan was to go fully static but we were using an unstable release of nuxt and so `target: static` was buggy.

Maybe this is something we should get sorted now that its a stable release? - before releasing lots of code that may or may not work with our preferred rendering method

MeganSteenkamp: @AndyClifford we have now added `target: static` and have tested it, seems to work as expected.

MeganSteenkamp: @AndyClifford Tom and I had a look at this and it appears as though our current config is correct.
The docs you linked are for Nuxt v2 🧡

AndyClifford: Sweet - okay looks like nuxt.com is the go to as opposed to nuxtjs.org

tom-blake: Yeah it's a little confusing 😵

tom-blake (approved):
Fantastic job @MeganSteenkamp - worth making sure that regular code blocks still work as part of your test plan 💪

AndyClifford (approved):

PR #550, opened 2022-12-21
🔧 Update build script to support guides directory from jekyll and nuxt

MeganSteenkamp:
We have existing pages [docs.centrapay.com/guides/*](https://docs.centrapay.com/guides/introduction) from our jekyll site. For the nuxt site, we want the URL https://docs.centrapay.com/guides/farmlands-pos-integration. 

Change to update the build script so that the guides directories can be merged ⛙

Test plan: Expect to be able to navigate to existing [/guides](https://docs.centrapay.com/guides/introduction) and [/guides/farmlands-pos-integration](https://docs.centrapay.com/guides/farmlands-pos-integration)


build.sh
```diff
@@ -11,8 +11,8 @@ cd ../
 
 yarn generate
 
-mv .output/public/api/_content _site/api
+rsync -a .output/public/api/_content _site/api
```
MeganSteenkamp, line 14: We can also add a `--remove-source-files` if we want to remove the files from the source.
Does it matter? What do we think? 🤨

tom-blake (approved):

PR #551, opened 2022-12-22
Add Farmlands card image 💳

MeganSteenkamp:
The Farmlands card image is used in the documentation to show the reader where the barcode and card number can be found.
- Add inline styling in markdown to dictate image height.
- Add styling for all images to dynamically resize images to stay inside a container, center align and have a margin at the bottom.


tailwind.config.js
```diff
@@ -5,6 +5,7 @@ const tailwindConfig = {
     './pages/**/*.{vue,js,css}',
     './assets/**/*.{vue,js,css}',
     './components/**/*.{vue,js,css}',
+    './content/**/*.md',
```
MeganSteenkamp, line 8: This tells Tailwind to look at our markdown files for Tailwind classes.
Ideally we should transform the files to HTML first: https://tailwindcss.com/docs/content-configuration#transforming-source-files although I'm having trouble with the configuration/ imports. 
This could be left as another task.

Another option is we just use inline styling for image heights and avoid tailwind processing these files as it's not really necessary.

MeganSteenkamp: Another option is to `safeList` the class we want so it doesn't get purged.

tom-blake (approved):

PR #553, opened 2022-12-23
Update Centrapay logo and text

Wulingoh:
The logo and text are aligned to top left. The user is able to click both logo and text to return to docs.centrapay.com.

Test plans:

1. Go to docs.centrapay.com 
2. Check the logo and text are aligned on the left side of the page
3. Go to docs.centrapay.com/v2/guides/farmlands-pos-integration
4. Click the logo and text to return to the homepage(docs.centrapay.com)

image

### DIFF
--- a/components/Header.vue
+++ b/components/Header.vue
@@ -1,10 +1,15 @@
 <template>
   <header class="w-full p-4 shadow-md sticky top-0 bg-surface-primary z-10">
-    <div class="prose desktop-gutters flex items-center justify-center space-x-2">
-      <icons-centrapay-logo class="w-12 h-12 fill-brand-accent" />
-      <h1 class="hidden md:block type-headline-3">
-        Centrapay Docs
-      </h1>
+    <div class="prose-a:no-underline max-w-7xl mx-auto flex items-center justify-start space-x-2">
+      <NuxtLink
+        to="/"
+        class="flex items-center"
+      >
+        <icons-centrapay-logo class="icon-3xl fill-brand-accent" />
+        <h1 class="hidden md:block type-headline-4 px-1">
+          Centrapay Docs
+        </h1>
+      </NuxtLink>
     </div>
   </header>
 </template>


components/Header.vue
```diff
@@ -1,8 +1,13 @@
...
-    <div class="desktop-gutters flex items-center justify-center space-x-2">
-      <icons-centrapay-logo class="w-12 h-12 fill-brand-accent" />
-      <span class="hidden md:block type-headline-3">Centrapay Docs</span>
+    <div class="flex items-center justify-start space-x-2">
+      <NuxtLink
+        to="/"
+        class="flex items-center"
+      >
+        <icons-centrapay-logo class="w-10 h-10 fill-brand-accent" />
+        <span class="hidden md:block type-headline-4 text-content-primary px-1">Centrapay Docs</span>
```
MeganSteenkamp, line 9: Can you please change this HTML tag to be <h1> ?
It is the main text for the heading and this makes a difference when people use screen readers 👍

components/Header.vue
```diff
@@ -1,8 +1,13 @@
...
   <header class="w-full p-4 shadow-md sticky top-0 bg-surface-primary z-10">
-    <div class="desktop-gutters flex items-center justify-center space-x-2">
-      <icons-centrapay-logo class="w-12 h-12 fill-brand-accent" />
-      <span class="hidden md:block type-headline-3">Centrapay Docs</span>
+    <div class="flex items-center justify-start space-x-2">
+      <NuxtLink
+        to="/"
+        class="flex items-center"
+      >
+        <icons-centrapay-logo class="w-10 h-10 fill-brand-accent" />
```
MeganSteenkamp, line 8: Instead of `w-10 h-10` you could use the class `icon-3xl` which has the same properties.
This means if we ever change our stylesheet everything will be consistent.

components/Header.vue
MeganSteenkamp: The class `text-content-primary` could also be removed as this is already in the definition for `type-headline-4` ☺️
You can see this in the file `assets/css/plugins/typography.js`.

MeganSteenkamp:
Nice work @Wulingoh 👏

MeganSteenkamp (approved):

tom-blake (approved):

PR #563, opened 2023-01-15
Remove createdByAccountId and createdByAccountName fields from create payment request operation fields list

bpal938:
createdByAccountId and createdByAccountName are no longer sourced from the create payment request body, instead using the internal jwt from styx 
https://www.notion.so/centrapay/a8cbc9d1738f484391ddb4f00ff7042f

This removes those values from the fields list for the endpoint.


IsaacMurtagh (commented):
lgtm

IsaacMurtagh (approved):

PR #566, opened 2023-01-16
update farmlands card image

AndyClifford:
The new image barcode is 9 digits instead of 12 digits


MeganSteenkamp (approved):

Ray2126 (approved):

Ray2126:
Non-blocking --> Is the image too big? Smaller images load quicker so we should only make them as big as they need to be

AndyClifford:
> Non-blocking --> Is the image too big? Smaller images load quicker so we should only make them as big as they need to be

Maybe - the file size is 185kb. I'll reduce it to match the size of the image that is being replaced 👍

AndyClifford:
accidentally closed this somehow

PR #571, opened 2023-01-17
Style links based on the type of link

Ray2126:
Designs based off https://www.figma.com/file/IgRlQWJTyGoZhlF4fVKD6B/Tailwind-Centrapay-Docs?node-id=2%3A2&t=PG8pI9FLaUsF6yv9-0

image

Test plan:
- Navigate to https://docs.centrapay.com/guides/farmlands-pos-integration/ and check that links look like the screenshot above

### DIFF
--- a/components/content/ProseA.vue
+++ b/components/content/ProseA.vue
@@ -1,0 +1,33 @@
+<template>
+  <NuxtLink
+    :to="props.href"
+    :target="isExternalLink ? '_blank' : '_self'"
+    class="no-underline hover:no-underline inline-flex justify-center items-center"
+  >
+    <span class="hover:underline"><slot /></span>
+    <span v-if="isExternalLink || isApiReferenceLink">&nbsp;</span>
+    <span
+      v-if="isApiReferenceLink"
+      class="rounded bg-interactive-quaternary h-5 w-10 inline-flex justify-center items-center"
+    >
+      <span class="text-content-inverse-primary text-xs">&nbsp;API&nbsp;</span>
+    </span>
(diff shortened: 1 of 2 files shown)


components/Prose.vue
```diff
@@ -6,7 +6,6 @@
       prose-h2:scroll-mt-16 prose-h2:mt-8 prose-h2:pt-8 prose-h2:border-t
       prose-h3:scroll-mt-20
       prose-lead:text-slate-500
-      prose-a:no-underline
```
MeganSteenkamp, line 9: There is a regression here with our markdown headings which are also links:
Screenshot 2023-01-18 at 9 21 24 AM

components/icons/LinkExternal.vue
```diff
@@ -0,0 +1,31 @@
+<template>
```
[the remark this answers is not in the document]
Ray2126: Yep, removed the component I added and replaced it with the existing one.

components/content/ProseA.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+  <NuxtLink
+    :to="props.href"
+    class="no-underline hover:no-underline inline-flex justify-center items-center"
```
MeganSteenkamp, line 4: We want an external link to open in a new tab so you'll need:
`:target = "isExternalLink ? '_blank' : '_self'"`

Ray2126: Nice catch! I've added this in

MeganSteenkamp (approved):

PR #572, opened 2023-01-17
Use Inter Font

tom-blake:
The Inter font is Centrapay's standard font that our designers use. Previously we weren't actually using it despite declaring it in our tailwind config. Now we're downloading it from the Google Font APIs

Test Plan:
- Ensure Inter font is being used throughout the site
- Ensure there are no visual regressions in existing pages

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -34,7 +34,11 @@ export default defineNuxtConfig({
     buildAssetsDir: '/nuxt/',
     head: {
       title: 'Centrapay Docs',
-      link: [{ rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' }],
+      link: [
+        { rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' },
+        { rel: 'preconnect', href: 'https://fonts.gstatic.com', crossorigin: 'anonymous'},
+        { rel: 'stylesheet', href: 'https://fonts.googleapis.com/css2?family=Inter:wght@300;400;500;600;700&display=swap', crossorigin: 'anonymous'},
+      ],
       meta: [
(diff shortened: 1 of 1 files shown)


tom-blake:
Farmlands Partner page without then with the Inter font:
image
image

tom-blake:
Farmlands POS Guide page without then with the Inter font:
image
image

nuxt.config.js
```diff
@@ -34,7 +34,11 @@ export default defineNuxtConfig({
     buildAssetsDir: '/nuxt/',
     head: {
       title: 'Centrapay Docs',
-      link: [{ rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' }],
+      link: [
+        { rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' },
+        { rel: 'preconnect', href: 'https://fonts.gstatic.com', crossorigin: 'anonymous'},
+        { rel: 'stylesheet', href: 'https://fonts.googleapis.com/css2?family=Inter:wght@300;400;500;600;700&display=swap', crossorigin: 'anonymous'},
```
MeganSteenkamp, line 40: I can see that in Tailwind example repos they have a `public/fonts/` directory with woff2 files stored that define the inter font - I'm not sure if this is a better way to do it 🤔

MeganSteenkamp (approved):

AndyClifford (approved):

nuxt.config.js
tom-blake: @MeganSteenkamp Yes there's trade-off to self-hosting (no bug fixes, you might not have the right support for certain browsers) or downloading from Google APIs (performance impact). I've chosen the latter as the performance hit was ok - we can tune this later if we need to 👍

PR #576, opened 2023-01-18
🐚 Fix Docs link in App Shell

MeganSteenkamp:
Link should be a `NuxtLink` as it is internal.
Route to docs homepage.
Change NuxtLink to have `to` property instead of `href`.

**Test plan**: Expect pressing `Docs` to route home. Expect other external links to open new tab.


components/AppShell.vue
```diff
@@ -63,11 +63,16 @@
                       :href="item.href"
                       target="_blank"
                       class="group mt-2 w-full flex items-center pl-2 pr-1 py-2 text-left text-content-primary text-base leading-6 font-medium rounded-md hover:bg-gray-200 focus:outline-none focus:ring-2 focus:ring-inset ring-focus-ring"
-                      :class="item.current ? 'bg-gray-50 text-content-primary': ''"
```
MeganSteenkamp, line 66: This is unneeded as none of the external links will be current.

AndyClifford (approved):

PR #577, opened 2023-01-18
Callout component

AndyClifford:
Callout component which is to be used in our markdown guides. We don't currently have a use case but when we migrate the rest of the content it will be.

We don't have a use case for the `heading` as seen in the figma design so it has been left out for now. This allows us to easily override the ProseBlockquote component.

**Figma Design:** https://www.figma.com/file/IgRlQWJTyGoZhlF4fVKD6B/Tailwind-Centrapay-Docs?node-id=501%3A5023&t=EjI1bg9db0kgpOau-0

Desktop
image

Mobile
image


AndyClifford:
If anyone has any suggestions of where a callout can be used with this change let me know and I will add it in - otherwise it can just sit there for now and we'll use it when we port the rest of the content across.

components/content/Callout.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <div class="bg-surface-secondary rounded-lg">
+    <div class="flex p-5 space-x-5">
+      <information-circle class="icon-md text-static-black" />
+      <div class="flex flex-col align space-y-1 text-sm leading-5">
```
MeganSteenkamp, line 5: I don't think `align` is a valid class 🤨

components/content/Callout.vue
```diff
@@ -0,0 +1,25 @@
...
+        <div
```
tom-blake, line 6: You probably stole this from Tailwind but could this be a `<p>` element instead of a `<div>`?

components/content/Callout.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <div class="bg-surface-secondary rounded-lg">
+    <div class="flex p-5 space-x-5">
```
tom-blake, line 3: Assuming you stole this from Tailwind but could we combine these two top-level divs to reduce nesting?

components/content/Callout.vue
```diff
@@ -0,0 +1,25 @@
...
+          {{ description }}
+        </p>
+      </div>
+    </div>
+  </div>
+</template>
+
+<script setup>
+const props = defineProps({
+  heading: { type: String, required: false, default: '' },
```
tom-blake, line 22: @AndyClifford FYI we don't actually have a requirement for a heading at this stage

components/content/Callout.vue
```diff
@@ -0,0 +1,25 @@
+<template>
```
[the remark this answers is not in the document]
AndyClifford: Given we don't have a requirement for headings yet I'll override the Blockquote component for now - It isn't much work to come back to this design in future 👍

MeganSteenkamp (approved):

PR #580, opened 2023-01-18
Make API badge smaller

AndyClifford:
Before:
image

After:
image

### DIFF
--- a/components/content/ProseA.vue
+++ b/components/content/ProseA.vue
@@ -8,9 +8,9 @@
     <span v-if="isExternalLink || isApiReferenceLink">&nbsp;</span>
     <span
       v-if="isApiReferenceLink"
-      class="rounded bg-interactive-quaternary h-5 w-10 inline-flex justify-center items-center"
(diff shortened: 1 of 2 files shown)


AndyClifford:
Do you think the text should also be smaller?

components/content/ProseA.vue
```diff
@@ -8,7 +8,7 @@
     <span v-if="isExternalLink || isApiReferenceLink">&nbsp;</span>
     <span
       v-if="isApiReferenceLink"
-      class="rounded bg-interactive-quaternary h-5 w-10 inline-flex justify-center items-center"
+      class="rounded-sm bg-interactive-quaternary h-3 w-6 inline-flex justify-center items-center"
```
MeganSteenkamp, line 11: Instead of a fixed height and width you can just remove this and it looks quite good. IMO this doesn't have enough spacing between the text and the top.

AndyClifford: @MeganSteenkamp I've made the text smaller which has increased this gap - have a look at the updated screenshot above. Do you think the gap should be even bigger?

Ray2126 (approved):
LGTM

PR #585, opened 2023-01-19
Seperate navigation menu from page URL

Ray2126:
Doing this will allow us to move the pages around in the nav menu, without breaking shared links. Hard coded the navigation menu for now because dynamically creating it is hard and we need this in place by the time we show Mitre10.

Test plan:
- The navigation menu should show `Connections` --> `Farmlands` --> `Farmlands POS Integration Guide`
- Click on `Farmlands`
  - Should take us to https://docs.centrapay.com/connections/farmlands
  - The nav menu items `Connections` and `Farmlands` should be highlighted.
- Click on `Farmlands POS Integration Guide`
  - Should take us to https://docs.centrapay.com/guides/farmlands-pos-integration
  - The nav menu items `Connections` and `Farmlands POS Integration Guide` should be highlighted

### DIFF
--- a/components/SiteNavigation.vue
+++ b/components/SiteNavigation.vue
@@ -5,13 +5,14 @@
       :key="item.title"
     >
       <Disclosure
+        v-if="item.children"
         v-slot="{ open }"
         as="div"
-        :default-open="currentPath.startsWith(item._path)"
+        :default-open="currentPath.startsWith(item.path)"
       >
         <DisclosureButton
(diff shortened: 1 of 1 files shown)


tom-blake (approved):

components/SiteNavigation.vue
```diff
@@ -5,13 +5,14 @@
       :key="item.title"
     >
       <Disclosure
+        v-if="item.children"
         v-slot="{ open }"
         as="div"
-        :default-open="currentPath.startsWith(item._path)"
+        :default-open="currentPath.startsWith(item.path)"
```
MeganSteenkamp, line 11: @Ray2126 @tom-blake FYI I there is a bug here.
When you go to https://docs.centrapay.com/guides/ the button for the Disclosure is facing down like it is open but it is not open.
Screenshot 2023-01-20 at 11 23 57 AM
farmlands-pos-integration/

PR #616, opened 2023-01-24
Make POS guide the default guides content

tom-blake:
Update the "Guides" home page card and the "Merchant Integrations" sidebar item to point to the Point of Sale guide. This is likely temporary until we have appropriate default content. The POS guide is a more sensible default in the meantime.

Test plan:
- Confirm that the Guides home page card takes readers to the POS guide content
- Confirm that the Merchant Integrations sidebar item takes readers to the Point of Sale guide content

### DIFF
--- a/components/SiteNavigation.vue
+++ b/components/SiteNavigation.vue
@@ -97,7 +97,7 @@ const navigation = [
     children: [
       {
         title: 'Merchant Integrations',
-        path: '/guides/merchant-integrations',
+        path: '/guides/point-of-sale',
         children: [
           {
             title: 'Point of Sale',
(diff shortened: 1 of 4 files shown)


components/SiteNavigation.vue
```diff
@@ -97,7 +97,7 @@ const navigation = [
     children: [
       {
         title: 'Merchant Integrations',
-        path: '/guides/merchant-integrations',
+        path: '/guides/point-of-sale',
```
MeganSteenkamp, line 100: This causes both items in the nav bar to be shown as active, non-blocking for me.
Screenshot 2023-01-25 at 8 33 14 AM

MeganSteenkamp (approved):

components/SiteNavigation.vue
tom-blake: Yep - I've added a task to fix this later on

PR #620, opened 2023-01-24
🔧 Reduce Heading length for Navigation and Guides

MeganSteenkamp:
Some of the existing headings are long (> 3-4 words) which makes the Table of Contents difficult to read. Change to reduce heading lengths where possible.

### DIFF
--- a/components/SiteNavigation.vue
+++ b/components/SiteNavigation.vue
@@ -116,7 +116,7 @@ const navigation = [
             path: '/guides/merchant-integration-barcode-flow',
           },
           {
-            title: 'Merchant Integration Error Handling',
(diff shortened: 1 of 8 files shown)


components/SiteNavigation.vue
```diff
@@ -157,7 +157,7 @@ const navigation = [
         path: '/connections/farmlands',
         children: [
           {
-            title: 'Farmlands POS Integration Guide',
+            title: 'POS Integration Guide',
```
MeganSteenkamp, line 160: Could maybe just be `Integration Guide` 🤨

tom-blake (approved):

PR #624, opened 2023-01-25
Update social media link previews for Farmlands pages

Ray2126:
Test plan:

- Share the Farmlands pages on Slack and Notion - see that they appear as expected
- Check that https://socialsharepreview.com/?url=https://docs.centrapay.com/connections/farmlands appears as expected for Facebook, LinkedIn and Twitter
- Check that https://socialsharepreview.com/?url=https://docs.centrapay.com/guides/farmlands-pos-integration appears as expected for Facebook, LinkedIn and Twitter
- Check that the title on https://docs.centrapay.com is "Centrapay Docs"
- Check that every other page has ` - Centrapay Docs` appended to the end of the title
- Regression: Check that the Farmlands landing page still has the banner image


Ray2126:
WIP: Code is all there but pending the content. https://centrapay.slack.com/archives/C0330QJ7M4H/p1674615796843979

Ray2126:
Taken out of WIP, Matthias is happy with it: https://centrapay.slack.com/archives/C0330QJ7M4H/p1674615796843979

Ray2126:
The social medias all have subtly different link previews but they should all look something like this:

image

image

pages/connections/farmlands.vue
```diff
@@ -103,9 +103,28 @@ const externalNavigation = [
...
+      property: 'og:image',
+      content: '/countryside.jpg',
+    },
+    {
+      property: 'og:image:width',
+      content: '1200'
+    },
+    {
+      property: 'og:image:height',
+      content: '630'
```
Ray2126, line 122: This is the recommended image size for OG - https://www.kapwing.com/resources/what-is-an-og-image-make-and-format-og-images-for-your-blog-or-webpage/#:~:text=OG%20Images%20are%20the%20first,ratio%20of%201.91%3A1).

pages/connections/farmlands.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div>
     <header
-      class="not-prose bg-bottom bg-cover bg-[#174B33] bg-blend-overlay bg-[url('~/assets/images/countryside.jpg')]"
+      class="not-prose bg-bottom bg-cover bg-[#174B33] bg-blend-overlay bg-[url('/countryside.jpg')]"
```
MeganSteenkamp, line 4: @Ray2126 was there a reason for moving this?
I think @tom-blake had a preference for this to be in the assets folder so that it can be processed by webpack to reduce the size of the content being served.

Ray2126: Yeah I couldn't get the image to show up in the link preview when it was in the `assets` folder. I could duplicate the image to both folders?

MeganSteenkamp: I think public is ok but check with Tom 👍

MeganSteenkamp (approved):

pages/connections/farmlands.vue
Ray2126: Checked with Tom and he is ok with public 👍

pages/connections/farmlands.vue
```diff
@@ -103,9 +103,28 @@ const externalNavigation = [
 ];
 
 useHead({
-  title: 'Connections - Farmlands',
+  title: 'Farmlands - Centrapay Docs',
```
tom-blake, line 106: We could suffix every page title with ` - Centrapay Docs`

I think there's a way to do this in the Nuxt Config

tom-blake, line 106: We could suffix every page title with ` - Centrapay Docs`

I think there's a way to do this in the Nuxt Config

tom-blake (approved):

pages/connections/farmlands.vue
Ray2126: Yup, I've done this - check out `app.vue`

PR #626, opened 2023-01-26
🐛 Fix: Open disclosure of active navigation item when link in body is clicked

MeganSteenkamp:
Previous behaviour: If you were on the Farmlands POS Integration guide and clicked a link in the body to a page under 'References', the 'References' section would show as active but the disclosure would not open. Change to open the disclosure of the active navigation section when the route changes.

Test plan: Expect disclosure of active navigation section to open when route changes.

Screenshot 2023-01-26 at 4 21 39 PM

### DIFF
--- a/components/SiteNavigationDisclosure.vue
+++ b/components/SiteNavigationDisclosure.vue
@@ -1,0 +1,100 @@
+<template>
+  <Disclosure
+    as="div"
+  >
+    <DisclosureButton
+      class="group mt-2 w-full flex items-center pl-3 pr-1 py-2 space-x-3 text-left rounded-md hover:bg-gray-200 focus:outline-none focus:ring-2 focus:ring-inset ring-focus-ring"
+      :class="currentPath.startsWith(navigationItem.path) ? 'bg-gray-100' : ''"
+      @click="open = !open"
+    >
+      <component
+        :is="navigationItem.icon"
+        class="icon-md flex-shrink-0 text-content-tertiary group-hover:text-content-secondary"
+        aria-hidden="true"
+      />
+      <span class="flex-1 text-content-primary text-base leading-5 font-medium">{{ navigationItem.title }}</span>
+      <disclosure-arrow-right :class="[open ? 'text-gray-400 rotate-90' : 'text-gray-300', 'ml-3 icon-md flex-shrink-0 transform transition-colors duration-150 ease-in-out group-hover:text-gray-400']" />
+    </DisclosureButton>
+    <transition
+      enter-active-class="transition duration-100 ease-out"
+      enter-from-class="transform scale-95 opacity-0"
+      enter-to-class="transform scale-100 opacity-100"
+      leave-active-class="transition duration-75 ease-out"
+      leave-from-class="transform scale-100 opacity-100"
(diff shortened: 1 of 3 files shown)


components/SiteNavigationDisclosure.vue
```diff
@@ -0,0 +1,101 @@
...
+
+import {
+  Disclosure,
+  DisclosureButton,
+  DisclosurePanel,
+} from '@headlessui/vue';
+
+const props = defineProps({
+  navigationItem: { type: Object, required: true },
+  urlToActiveNav: { type: Object, required: true },
```
MeganSteenkamp, line 72: I decided to pass this through as a prop so the manually configured nav is all in one place.

components/SiteNavigationDisclosure.vue
```diff
@@ -0,0 +1,101 @@
...
+
+const emit = defineEmits([ 'link-clicked' ]);
+
+const route = useRoute();
+const currentPath = computed(() => {
+  const path = route.path.endsWith('/') ? route.path.slice(0, -1) : route.path;
+  return props.urlToActiveNav[path] || path;
+});
+const open = ref(currentPath.value.startsWith(props.navigationItem.path));
+watch(() => route.path, () => {
```
tom-blake, line 83: This `watch` is difficult to understand - can we watch the `currentPath` property directly instead to make this cleaner? 😄

tom-blake, line 83: @MeganSteenkamp  This watcher is difficult to understand - can we watch the currentPath property directly instead to make this cleaner? 😄

tom-blake (changes requested):

components/SiteNavigationDisclosure.vue
MeganSteenkamp: @tom-blake watching `currentPath` doesn't work. 
From the Vue docs: "A computed property will only re-evaluate when some of its reactive dependencies have changed." My understanding is that clicking a NuxtLink doesn't directly affect anything within this component, therefore navigating doesn't cause `currentPath` to be re-evaluated. Therefore, the watch function doesn't execute. This was the problem with the initial code.
See: https://vuejs.org/guide/essentials/computed.html#computed-caching-vs-methods

tom-blake: @MeganSteenkamp Hmm I'd expect that clicking a NuxtLink would update `route.path` used by the computed property and cause it to re-evaluate 🤔

tom-blake: @MeganSteenkamp This seems to work locally unless I'm missing something?
```
const open = ref(currentPath.value.startsWith(props.navigationItem.path));
watch(currentPath, (newCurrentPath) => {
  if (open.value) {
    // If disclosure is already open, leave it open
    return;
  }
  open.value = newCurrentPath.startsWith(props.navigationItem.path);
});
```

MeganSteenkamp: Thanks @tom-blake - can't remember what I tried but it was obviously incorrect 👀 
I have updated this now 🧡

tom-blake (approved):

PR #633, opened 2023-01-30
Make navigation dynamically pick up pages in the content folder

Ray2126:
We cut some corners by making the navigation static to hit some deadlines.

Having a static navigation menu will mean that when adding pages, there will need to be additional steps followed to add the page to the left hand navigation menu on the site.

Making the navigation menu dynamic means that when adding pages, you will only have to define where it should sit in the nav. The new page will automatically appear on the navigation menu on the site.

Adding new sections will still need to be manual. This is a fine compromise because new sections are rare.

Test plan:
- Ensure that there are no regressions in the site
- Add a new page with a navPath --> this page should automatically appear in the nav menu on the site

### DIFF
--- a/components/SiteNavigation.vue
+++ b/components/SiteNavigation.vue
@@ -1,12 +1,11 @@
 <template>
   <div>
     <div
-      v-for="item in navigation"
+      v-for="item in menu.children"
       :key="item.title"
     >
-      <div v-if="item.children">
+      <div v-if="item.children.length">
         <SiteNavigationDisclosure
-          :url-to-active-nav="urlToActiveNav"
           :navigation-item="item"
           @link-clicked="$emit('link-clicked')"
         />
@@ -14,7 +13,7 @@
       <NuxtLink
         v-else
         target="_blank"
-        :to="item.path"
+        :to="item.to"
         class="group mt-2 w-full flex items-center pl-3 pr-1 py-2 space-x-3 text-left rounded-md hover:bg-gray-200 focus:outline-none focus:ring-2 focus:ring-inset ring-focus-ring"
       >
         <component
@@ -29,104 +28,8 @@
 </template>
 
 <script setup>
+import Navigation from '../utils/Navigation';
 const emit = defineEmits([ 'link-clicked' ]);
 
-const config = useRuntimeConfig();
-// TODO Make the navigation dynamically pick up all the pages in content folder
-const navigation = [
-  {
-    title: 'Reference',
-    path: '/reference',
-    icon: 'receipt',
-    children: [
-      {
-        title: 'Merchant Integrations',
-        path: '/guides/point-of-sale',
-        children: [
-          {
-            title: 'Point of Sale',
-            path: '/guides/point-of-sale',
-          },
-          {
-            title: 'Requesting Payment',
-            path: '/guides/requesting-payment',
-          },
-          {
-            title: 'QR Code Flow',
-            path: '/guides/merchant-integration-qr-code-flow',
-          },
-          {
-            title: 'Barcode Flow',
-            path: '/guides/merchant-integration-barcode-flow',
-          },
-          {
-            title: 'Error Handling',
-            path: '/guides/merchant-integration-error-handling',
-          },
-          {
-            title: 'Transaction Reporting',
-            path: '/guides/transaction-reporting',
-          },
-          {
-            title: 'Initiating Refunds',
-            path: '/guides/initiating-refunds',
-          },
-          {
-            title: 'Line Items',
-            path: '/guides/line-items',
-          },
-          {
-            title: 'Payment Conditions',
-            path: '/guides/payment-conditions',
-          },
-          {
-            title: 'Requesting Pre Auth',
-            path: '/guides/requesting-pre-auth',
-          },
-          {
-            title: 'Patron Not Present',
-            path: '/guides/patron-not-present',
-          },
-        ]
-      }
-    ]
-  },
-  {
-    title: 'Connections',
-    path: '/connections',
-    icon: 'connections',
-    children: [
-      {
-        title: 'Farmlands',
-        path: '/connections/farmlands',
-        children: [
-          {
-            title: 'POS Integration Guide',
-            path: '/guides/farmlands-pos-integration',
-          }
-        ]
-      }
-    ]
-  },
-  {
-    title: 'API',
-    path: config.baseUrl + '/api',
-    icon: 'settings',
-  }
-];
-
-const urlToActiveNav = {
-  '/guides/farmlands-pos-integration': '/connections/farmlands/farmlands-pos-integration',
-  '/guides/payment-conditions': '/reference/merchant-integrations/payment-conditions',
-  '/guides/requesting-pre-auth': '/reference/merchant-integrations/requesting-pre-auth',
-  '/guides/patron-not-present': '/reference/merchant-integrations/patron-not-present',
-  '/guides/merchant-integration-error-handling': '/reference/merchant-integrations/merchant-integration-error-handling',
-  '/guides/line-items': '/reference/merchant-integrations/line-items',
-  '/guides/initiating-refunds': '/reference/merchant-integrations/initiating-refunds',
-  '/guides/point-of-sale': '/reference/merchant-integrations/point-of-sale',
-  '/guides/merchant-integration-qr-code-flow': '/reference/merchant-integrations/merchant-integration-qr-code-flow',
-  '/guides/transaction-reporting': '/reference/merchant-integrations/transaction-reporting',
-  '/guides/requesting-payment': '/reference/merchant-integrations/requesting-payment',
-  '/guides/merchant-integration-barcode-flow': '/reference/merchant-integrations/merchant-integration-barcode-flow',
-};
+const menu = Navigation.getMenu();
 </script>
--- a/content/guides/farmlands-pos-integration.md
+++ b/content/guides/farmlands-pos-integration.md
@@ -2,6 +2,10 @@
 title: Farmlands POS Integration Guide
 img: /farmlands-pos-integration-cover.jpg
 description: Centrapay and Farmlands have entered into a partnership to allow Farmlands Card Partners to accept Farmlands Card payments at the point of sale.
+nav:
+  path: Connections/Farmlands
+  title: POS Integration Guide
+  order: 1
 ---
 
 Centrapay and Farmlands have entered into a partnership to allow Farmlands Card Partners to accept Farmlands Card payments at the point of sale. The benefits include real-time standard industry card checks, the elimination of reconciliation EDI files or manual PDF invoicing, and it unlocks the ability to accept many more payment methods on the Centrapay platform. This solution positions Card Partners to be able to accept digital Farmlands Card payments in the future.
--- a/content/guides/requesting-payment.md
+++ b/content/guides/requesting-payment.md
@@ -1,5 +1,8 @@
 ---
 title: Requesting Payment
+nav:
+  path: Reference/Merchant Integrations
+  order: 2
 ---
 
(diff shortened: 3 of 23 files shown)


content/guides/farmlands-pos-integration.md
```diff
@@ -1,7 +1,10 @@
 ---
-title: Farmlands POS Integration Guide
+title: POS Integration Guide
```
Ray2126, line 2: The title on this page (Farmlands POS Integration Guide) was different to the title on the nav (POS Integration Guide).

These two titles should match to avoid confusion for the user. Changing the title on the nav to "Farmlands POS Integration Guide" causes the title to wrap which doesn't look very nice.

MeganSteenkamp: @Ray2126 we had a few instances like this so it may be best to go through them with Tom and check whether we want to rename anything else.
Another example is that `Merchant Integration Error Handling` was just `Error Handling` in the nav.

utils/toTitleCase.js
```diff
@@ -0,0 +1,12 @@
+// Taken from https://stackoverflow.com/a/27441234
+function toTitleCase(str){
```
MeganSteenkamp, line 2: I'm not sure that we need all this functionality - we likely only need a function that can split `lisp-case` words. 
Tailwind has text transformation that can capitalise the words for you: https://tailwindcss.com/docs/text-transform

MeganSteenkamp:
Great job @Ray2126, this is looking good 🧡

content/guides/farmlands-pos-integration.md
Ray2126: Good catch!

image

@tom-blake Above is a screenshot with the local nav on the left and the production nav on the right. I think we can rename them all so:
- QR Code Flow for Merchants --> QR Code Flow
- Barcode Flow for Merchants --> Barcode Flow
- Merchant Integration Error Handling --> Error Handling

This change will mean the titles on the respective pages will also change. I think this is okay because they all come under the `Merchant Integrations` section.

Ray2126: The missing context will also be remedied once we add breadcrumbs.

utils/toTitleCase.js
Ray2126: Sure, we can change it to only handle `lisp-case`. I like the function capitalising the words though because it is unit tested whereas the `SiteNavigation` is not.

components/SiteNavigation.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div>
     <div
-      v-for="item in navigation"
+      v-for="item in menu"
       :key="item.title"
     >
       <Disclosure
```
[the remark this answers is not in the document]
Ray2126: Yep, I've checked for these regressions and it looks all good :)

content/guides/farmlands-pos-integration.md
Ray2126: I've had a chat with Tom and we decided to put a Nav title in the front matter. This way, the nav and the page title is decoupled. The nav will default to the page title if the Nav title does not exist.

package.json
```diff
@@ -22,6 +26,7 @@
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-standard": "^4.0.0",
     "eslint-plugin-vue": "^7.9.0",
+    "jest": "^29.4.1",
```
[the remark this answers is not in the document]
Ray2126: Yes, Vitest is cleaner! I wasn't sure if we wanted to add a new tool here but it seems Vitest is compatible with Jest syntax out of the box.

components/SiteNavigation.vue
```diff
@@ -32,101 +32,6 @@
...
-  '/guides/line-items': '/reference/merchant-integrations/line-items',
-  '/guides/initiating-refunds': '/reference/merchant-integrations/initiating-refunds',
-  '/guides/point-of-sale': '/reference/merchant-integrations/point-of-sale',
-  '/guides/merchant-integration-qr-code-flow': '/reference/merchant-integrations/merchant-integration-qr-code-flow',
-  '/guides/transaction-reporting': '/reference/merchant-integrations/transaction-reporting',
-  '/guides/requesting-payment': '/reference/merchant-integrations/requesting-payment',
-  '/guides/merchant-integration-barcode-flow': '/reference/merchant-integrations/merchant-integration-barcode-flow',
-};
+const content = await queryContent().find();
+const { menu, pathToActiveNav } = navigation.create({ baseUrl: config.baseUrl, content });
```
tom-blake, line 36: It looks like `pathToActiveNav ` is not inside this component directly. I'm wondering if we should decouple building the data structures from retrieving them so that different components can pull in what they need?
-  `SiteNavigation.vue` uses the `menu data structure
- `SiteNavigationDisclosure.vue` uses the `pathToActiveNav` data structure

Ray2126: Yup, I've created separate getters for the `menu` data structure and the `pathToActiveNav` data structure so the components can just get what they need.

utils/NavigationItem.js
```diff
@@ -0,0 +1,27 @@
+const dfs = function ({ navigationItem, targetTitle }) {
```
MeganSteenkamp, line 1: Could this be defined as a normal function i.e. `function dfs() {}`?
I'm not sure on the naming, I am guessing it stands for depth first search but I'm not sure that's obvious.

MeganSteenkamp (approved):

utils/navigation.js
```diff
@@ -0,0 +1,79 @@
+import NavigationItem from './NavigationItem';
+import Page from './Page';
+
+function buildMenu({ navigationItem, pathToActiveNav, page, depth }) {
+  const hasChildren = page.hasChildrenAtNavDepth(depth);
+  const titleAtDepth = page.titleAtNavDepth(depth);
+  let item = navigationItem.findItemByTitle(titleAtDepth);
```
tom-blake, line 7: @Ray2126 There's a bit of indirection here where we're doing a search outside the class-then-insert. I'd find it more intuitive to have a `Navigation.insert` method that encapsulates searching for the right position AND inserts the `NavigationItem` - what do you think?

tom-blake: While it's inserting `NavigationItems` the `insert` method could just insert any intermediate nodes that don't exist yet on the way down

tom-blake (changes requested):

utils/navigation.js
```diff
@@ -0,0 +1,79 @@
...
+        new NavigationItem({
+          title: 'API',
+          to: baseUrl + '/api',
+          icon: 'settings',
+        }),
+      ]
+    });
+    this.pathToActiveNav = {};
+    const pages = content
+      .map(c => Page.fromContent(c))
```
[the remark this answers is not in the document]
Ray2126: I like the distinction.

`Page` - an encapsulation of all the metadata on a page in the content directory. This includes the front matter. We could expand this class to handle things like the breadcrumbs e.g. `page.toBreadcrumbs`.

`NavigationItem` - An instance of an item in the nav menu i.e. a node in the nav menu tree.

If we were to merge them, the `nav` property from the page front matter would be on the `NavigationItem` class. This doesn't make sense to me.

utils/NavigationItem.js
Ray2126: Yep for sure, can rename and turn into a normal function!

utils/navigation.js
```diff
@@ -0,0 +1,79 @@
...
+    });
+    // FIXME remove this block once merchant integrations has a landing page
+    const merchantIntegrations = this.menu.findItemByTitle('Merchant Integrations');
+    if(merchantIntegrations) {
+      merchantIntegrations.to = merchantIntegrations.children[0].to;
+    }
+  }
+}
+
+let navigation;
```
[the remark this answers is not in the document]
Ray2126: Are you thinking to make the methods static on the `Navigation` class? If not, where would you store the instance of the `Navigation` class?

utils/navigation.js
Ray2126: I don't understand what you mean. Can we catch up about this?

content/guides/farmlands-pos-integration.md
```diff
@@ -2,6 +2,10 @@
 title: Farmlands POS Integration Guide
 img: /farmlands-pos-integration-cover.jpg
 description: Centrapay and Farmlands have entered into a partnership to allow Farmlands Card Partners to accept Farmlands Card payments at the point of sale.
+nav:
+  path: /connections/farmlands/farmlands-pos-integration
```
Neeraj-Ghodla, line 6: Shouldn't this path be `/connections/farmlands/pos-integration-guide`

utils/navigation.js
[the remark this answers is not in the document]
Ray2126: Caught up with Tom and we decided to keep the `Page` class but remove the `NavigationItem` class.

`NavigationItem` is a dumb object that doesn't need to be it's own class. `Page` is a nice abstraction and can be expanded further for things like breadcrumbs, SEO metadata, and anything else we would want to put in the front matter.

utils/navigation.js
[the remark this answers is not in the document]
Ray2126: Caught up with Tom. We will make the methods below static on the `Navigation` class, and export the `Navigation` class from this module.

utils/navigation.js
Ray2126: Caught up with Tom, we will add an `insert` method on the `Navigation` class.

content/guides/farmlands-pos-integration.md
Ray2126: Hmm the last part of the nav path actually doesn't matter. This is because once we get to the last part, and we see that there are no more children, we set the title of the navigation menu to the `title` or `nav.title` in the front matter.

I will rework the solution so that we don't need to define that last part i.e. `/connections/farmlands`. I think this makes more sense as it is saying this page belongs to the `Connections` section under the `Farmlands` sub-section.

utils/__tests__/Navigation.test.js
```diff
@@ -0,0 +1,137 @@
...
+      const result = Navigation.getCurrentNavPath('/guides/farmlands-pos-integration');
+      expect(result).toEqual('/connections/farmlands/farmlands-pos-integration');
+    });
+
+    it('should not map current route to nav path when the current route does not have to be mapped', () => {
+      Navigation.create({
+        baseUrl: 'https://www.baseurl',
+        content: [
+          {
+            _path: '/connections/farmlands/farmlands-pos-integration',
```
tom-blake, line 124: It was hard to understand what this test was testing - if we're just testing the default return value for `getCurrentNavPath`, perhaps this could be simpler/clearer?

tom-blake (approved):

utils/__tests__/Navigation.test.js
Ray2126: I have removed that test, it is not testing anything that the test wasn't testing.

content/guides/requesting-payment.md
```diff
@@ -1,5 +1,8 @@
 ---
 title: Requesting Payment
+nav:
+  path: /reference/merchant-integrations
```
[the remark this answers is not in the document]
Ray2126: I have changed the format of the nav path.

I couldn't remove the `toKebabCase` function though because we still need a format of the nav path that can be compared to the `to` property on nav items. This is so we can highlight the parent navigation items. E.g.

Nav path: `References/Merchant Integrations`

So the `References` section and the `Merchant Integrations` section need to be highlighted. This is done by using the `to` property on the navigation item which are `/references` and `/references/merchant-integrations` respectively.

We need to use a `startsWith` syntax to make sure both sections are highlighted so we need to transform the nav path into a route-style path --> `References/MerchantIntegrations --> /references/merchant-integrations`

PR #650, opened 2023-02-10
Add test onboarding status to merchants and businesses

Ray2126:


Ray2126:
Test will no longer be an onboarding status

PR #655, opened 2023-02-17
🤝 Fix Lighthouse accessibility issues for buttons, links, and images

MeganSteenkamp:
Lighthouse identified issues with buttons and links including:
* Buttons not having an accessible name
* Links not having a discernable name
* Images having no alt text

Test plan: Test the home page, Farmlands Partner page, Farmlands POS Guide and a Reference with Lighthouse. Expect no errors with buttons, links and images.

NB There are outstanding accessiblity issues with colour contrast that have been raised with the design team.

### DIFF
--- a/src/components/Farmlands.vue
+++ b/src/components/Farmlands.vue
@@ -42,6 +42,7 @@
             target="_blank"
             class="icon-md text-interactive-tertiary hover:text-interactive-tertiary-active"
           >
+            <span class="sr-only">{{ link.description }}</span>
             <component :is="link.logo" />
           </a>
         </div>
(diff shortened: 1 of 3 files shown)


tom-blake (approved):

src/components/Farmlands.vue
```diff
@@ -59,7 +60,10 @@
           Your business can now either fully integrate your retail point of sale system or access a secure web portal when accepting Farmlands Card payments anytime, anywhere.
         </p>
         <div class="m-auto max-w-[450px] pt-4 pb-3">
-          <img src="/farmlands-flow.png">
+          <img
+            src="/farmlands-flow.png"
+            alt=""
```
Ray2126, line 65: Is an empty alt text the same as having alt text in terms of accessibility?

MeganSteenkamp: @Ray2126 if you supply no alt text, this would read as “slash-farmlands-dash-flow-dot-p-n-g, Unlabelled image” which isn't helpful. I didn't think the image added any context that wasn't written on the page and didn't have an alt description that would add any value.

Adding an empty alt tag skips the image being announced at all on a screen reader.

PR #662, opened 2023-02-22
Add farmlands portal introduction guide

IsaacMurtagh:


src/content/guides/farmlands-portal-introduction.mdx
```diff
@@ -0,0 +1,200 @@
+---
+title: Farmlands Portal Introduction Guide
+description: Farmlands Co-operative (Farmands) have partnered with Centrapay Payment Gateway (Centrapay) to deliver a secure web portal for Farmlands Card payments.
+nav:
+  path: Connections/Farmlands
+  title: Portal Introduction Guide
+  order: 2
```
tom-blake, line 7: @IsaacMurtagh Worth checking in with @laurencemcknight on this order value - I don't know what will happen if 2 pages have the same order number right now. I don't think we have guardrails in place to detect this at build time (yet)

src/content/guides/farmlands-portal-introduction.mdx
```diff
@@ -0,0 +1,200 @@
...
+| Check ID               | This response may occur for a high-value sale. This means the transaction is pending authorisation until the Card Partner confirms they have authenticated the customer’s photo ID against the name on the Farmlands Card. |
+| Declined               | The authorisation is declined due to an invalid card number, a credit limit exceeded, or Card Partner is unable to verify a high-value transaction. Refer to Farmlands Contact Centre 0800 200 600                         |
+| Release                | The funds “on hold” are cancelled by Card Partner. This may occur when a Card Partner is unable to fulfil an order. (This is not used for a refund).                                                                       |
+
+### Create a Farmlands Authorisation
+
+1. Select Create Authorisation Request from the main dashboard.
+
+    <img src="/farmlands-payment-create-authorization-request.png" alt="Create a Farmlands authorisation request on the Centrapay business portal" style=""/>
```
tom-blake, line 64: You should be able to use Markdown syntax for these images rather than embedding HTML here

`Create a Farmlands authorisation request on the Centrapay business portal`

We aimed to keep our content as close to raw markdown as possible. In the future when there's a CMS on top of our site, we want others to be able to update images (although admittedly I'm not entirely sure how that would work as we haven't selected a CMS yet)

src/content/guides/farmlands-portal-introduction.mdx
```diff
@@ -0,0 +1,200 @@
...
+
+Farmlands require this “approved” Short Code to be shown on your invoice sent for payment. The Short Code should be copied (by clicking on the copy button shown) and pasted into the PO field or the current field where you would normally insert the Farmlands Card Number.
+
+<img src="/farmlands-payment-authorisation-approved.png" alt="Payment Authorization approved on the Centrapay business portal" style=""/>
+
+Using the authorisation portal provides more assurance for our Card Partners to receive payment on time. Invoicing terms and conditions still apply e.g. invoices are to be in on the first working day before 11.59 pm. Farmlands encourages invoices to be submitted daily rather than all at the end of the month. This allows Farmlands to apply the purchase to the customer’s account in a timely manner.
+
+Invoices must include the “short code” on every invoice. This should be in the same field for every invoice. The Farmlands Card Number can be provided as an addition to the short code in another field.
+
+> **Instore Transactions** <br />
```
tom-blake, line 108: Not a fan of putting these `<br />` elements in our markdown 🤔 

Do we need a header on these callouts at all? Or does tabbing the content in 1 or 2 keep the body within the callout?

IsaacMurtagh: No that doesn't seem to keep the content inside the callout

src/content/guides/farmlands-portal-introduction.mdx
IsaacMurtagh: I can keep it close to markdown. However, I think you'll find where ever markdown is accepted so is HTML.
Markdown is easier to read and so it's preferred for content.

When you use a CMS images are hosted somewhere then in either the build step or server-side rendering you retrieve that image from the server.

src/content/guides/farmlands-portal-introduction.mdx
```diff
@@ -0,0 +1,200 @@
+---
+title: Farmlands Portal Introduction Guide
+description: Farmlands Co-operative (Farmands) have partnered with Centrapay Payment Gateway (Centrapay) to deliver a secure web portal for Farmlands Card payments.
```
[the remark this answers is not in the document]
IsaacMurtagh: No they don't, not yet

src/content/guides/farmlands-portal-introduction.mdx
IsaacMurtagh: Nice spot, I don't think it does anything, setting them both to one, but i'll set it to 3 for now.

src/content/guides/farmlands-portal-introduction.mdx
tom-blake: Damn.

We could move these callouts under the correct `Create a Farmlands Authorisation` steps above so that we don't need the headers at all. Happy for this to be considered at a later time though - let's publish this

tom-blake (approved):

PR #675, opened 2023-02-28
Add accordion component

laurencemcknight:
This PR adds an expandable accordion component. 
This component displays a title. When the title is clicked a description is displayed underneath.
Farmlands have requested this to be used for the FAQ section in the Farmlands guides.

Testing:
Assert that you can tab between accordions
Assert that you can use enter to expand an accordion
Assert that content appears correctly on different screen sizes and in different resolutions

image


tom-blake (changes requested):

tom-blake:
@laurencemcknight This PR needs a test plan

We should also get a design review if you haven't already. IMO the FaQs are too close together and I'm not sure there's much benefit in making the Question text bold

src/content/guides/farmlands-portal.mdx
```diff
@@ -195,55 +197,69 @@ On the Settings page, you can change your business information. Please note this
...
+</AccordionItem>
+
+<AccordionItem client:visible style="border-bottom-width:1px">
+  <span slot="title">**What do we do if the customer cancels their order <u>before</u> we have sent the invoice to Farmlands?**</span>
+  <span slot="description">If the entire order has been cancelled, release the hold on the customer’s funds via the Portal. There is no requirement to send an invoice to Farmlands.</span>
+</AccordionItem>
+
+<AccordionItem client:visible style="border-bottom-width:1px">
+  <span slot="title">**What do we do if we need to amend the value of the order prior to sending the invoice to Farmlands?**</span>
+  <span slot="description">
```
tom-blake, line 217: @laurencemcknight Interested in what the DOM looks like here - `<div>` might be more appropriate to wrap slot content than a span

laurencemcknight:
>

src/content/guides/farmlands-portal.mdx
laurencemcknight: I have removed these HTML elements, so I'm going to mark this as resolved.

tom-blake (commented):
@laurencemcknight After chatting I think you're on the right track here 💪 I'll remove my request for changes so that someone else can approve it if needed

src/content/guides/farmlands-portal.mdx
```diff
@@ -195,55 +197,51 @@ On the Settings page, you can change your business information. Please note this
...
-**What if the barcode will not scan or is obscured?**
-
-The 9-digit Card number can be entered manually. This is printed on the back below the barcode and on the front of the Card under the Cardholder's name.
-
-If neither is visible, please request another method of payment. Advise the Cardholder to contact Farmlands to request a new Card.
-
-**The authorisation request has prompted us to “Check ID.” The customer is not the actual Cardholder but one of the workers of the Cardholder. Can I accept the Card?**
-
-No, unfortunately you must only accept the Card if the true owner of the Card is making the purchase.
+<FAQ client:visible title="What do we do if a customer returns one or more items, and the invoice has already been sent to Farmlands?">
```
IsaacMurtagh, line 200: I'm not a fan of this, I think logically there needs to be a wrapping component for all of them.

This is how you might track state in the future to have one open at a time.
Or in your case, insert a line in-between disclosures.

Looking at what FAQ does, it's adding a single style and then passing the slot and props through.
I suspect there's concern about adding this style to the disclosure as we may want to re-use disclosure.
I think if that's the case - get rid of Dislocsure.vue and call it FaqDisclosure.vue, you don't need that extra layer of abstraction.

IsaacMurtagh (approved):
LGTM!

src/components/Disclosure.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <Disclosure>
+    <DisclosureButton
+      class="flex w-full flex-row justify-between px-0 my-4 focus:ring-1 ring-offset-8 rounded-md type-subtitle-2"
+      style="text-align:left"
```
IsaacMurtagh, line 5: There's a tailwind class for this, https://tailwindcss.com/docs/text-align#setting-the-text-alignment

PR #677, opened 2023-03-01
Migrate farmlands connection page to mdx

alexBinns:
What it says on the can


tom-blake:
@alexBinns Can you add a test plan and screenshots to this PR? 💪

There have been some spacing changes to `Layout.vue` and a slot added to the `BaseLayout` that will affect all the pages on our site. 

I just want to be confident that there haven't been any regressions and I'm wondering if the footprint of these changes could be smaller 🤔

tom-blake:
@alexBinns A couple of thoughts, sorry I can't comment directly on code lines on GitHub mobile. I've tried to explain my thinking as much as possible ♥️

- The `BaseLayout` component is intended to be everything that wraps the main content of a page (header, nav, footer). This layout is used on every page of the docs site. 
- I don't think we should add your header slot to `BaseLayout`. No changes should be needed here.
- We've encapsulated the presentation of the main page content with the `Layout` component (markdown content is injected here + a table of contents).
- Up until now, we've only had one type of doc -> a guide
- The Farmlands connection page has a different layout than our guides (bigger header, no table of contents). I think it's totally reasonable for connections pages to have a bespoke structure for each future partner. Gives us flexibility in the future.

To reduce the scope of these changes and prevent them affecting the entire site, we could:
- Rename `Layout.vue` to `Guide.vue` to more accurately reflect the type of doc it should be used for. This would be the only change to this file so we can be confident of no regressions.
- Use a different layout component specifically for Farmlands and just use that to receive the markdown content

src/components/Cards.vue
```diff
@@ -0,0 +1,24 @@
+<template>
```
IsaacMurtagh, line 1: I'm not a fan of having a collection component with a name like "Cards" If this is necessary I think you need to come up with a better name.

I think what this is trying to do is place components next to each other until the screen is too small.

What about a comopnent called Grid that does this for you.

The markdown might look like this.

```
<Grid md-cols="2">
  <Card>
  ...
  </Card>
  <Card>
  ...
  </Card>
</Grid>
```

Then the vue component looks like
```
<template>
 <div class="grid gap-6 lg:gap-12", :class="(something computed from properties)">
  <slot />
 </div>
</template>
```

src/components/Cards.vue
```diff
@@ -0,0 +1,24 @@
...
+
+<div class="not-prose flex flex-col md:flex-row pt-4 md:space-x-6 lg:space-x-12 pb-16 space-y-6 md:space-y-0">
+<span
+  v-for="card in cards">
+  <Card
```
IsaacMurtagh, line 6: Can you do a lot of this with slots? Card is a very generic name for a component so forcing all content to go through props means it's not flexible at all

IsaacMurtagh (changes requested):

src/components/Cards.vue
alexBinns: I was trying to reduce styling in the markdown file. I think changing the name is ok but I don't wholly agree with your suggestion. I will have a think about it. Also you can't pass more than one think into the slot

IsaacMurtagh: My Suggestion wouldn't include any styling in the markdown file

src/components/Cards.vue
alexBinns: I have updated to use slot

alexBinns:
@tom-blake I have reduced the changes to adding one conditional. I think there is already a lot in this PR I will do a followup PR with a new layout file for the connections.

src/components/Cards.vue
alexBinns: I have updated to your suggestions I haven't added the properties if someone wants to extend later they can

src/components/Grid.astro
```diff
@@ -0,0 +1,4 @@
+<div>
+<div class="not-prose flex flex-col md:flex-row pt-4 md:space-x-6 lg:space-x-12 pb-16 space-y-6 md:space-y-0">
```
IsaacMurtagh, line 2: You should use the grid classes for this, it's much easier to read and more true to what it's doing.
If there are 10 children, you don't want them all on the same line.

https://tailwindcss.com/docs/grid-template-columns

src/pages/connections/farmlands.mdx
```diff
@@ -0,0 +1,34 @@
...
+In partnership with Centrapay, Farmlands has two new digital solutions that enable Card Partners to authorise Card payments in real-time, and process invoices to Farmlands faster. There are no transactional costs for Card Partners using these solutions.
+
+Your business can either fully integrate the Farmlands solution to your retail point of sale system, or access a secure web portal when accepting Farmlands Card as a form of payment for a sale anytime, anywhere.
+
+![Farmlands card flow](/farmlands-flow.png)
+
+## Solutions
+<Grid>
+  <Card
+    title='1. POS API'
```
IsaacMurtagh, line 17: Why is everything passed in as a prop?

A card component is a well-defined component in UI. It's not opinionated on how the content is laid out or what can go in it.

It's more there to add padding, edging, shadowing etc.
Is there a way you can use a slot, or should this be a more specific component

src/layouts/ConnectionLayout.astro
```diff
@@ -0,0 +1,22 @@
...
+import GuideLayout from '../components/GuideLayout.vue';
+
+const img = 'countryside.jpg';
+const fullWidth = true;
+const headings = {};
+---
+
+<BaseLayout>
+  <Farmlands/>
+    <GuideLayout
```
IsaacMurtagh, line 13: I think adding variants to the guide is going to couple the display of a guide and cover pages together.
Was there a suggestion from Tom to break this out into its own layout?

src/pages/connections/farmlands.mdx
alexBinns: I'm reusing an existing component. I'm not against updating it but I would like to do it in a follow up PR

src/layouts/ConnectionLayout.astro
alexBinns: As per my comment to Tom I have added this small change and will create another layout in a follow up PR

src/components/Grid.astro
alexBinns: This has been implemented

IsaacMurtagh (approved):
Approving but record tasks to fix to
-  remove props from card component, and
- Remove variant prop from guide layout

PR #691, opened 2023-03-08
Add featured docs to home page

IsaacMurtagh:
image


src/pages/index.astro
```diff
@@ -1,8 +1,52 @@
...
+      <p class="type-body-2 max-w-[720px]">
+        For more help getting setup, contact us via email at
+        <a href="mailto:integrations@centrapay.com">
+          integrations@centrapay.com.
+        </a>
+      </p>
+      <h2 class="type-headline-2">Featured Docs</h2>
+    </Prose>
+      <div class="grid md:grid-cols-2 lg:grid-cols-3 gap-4 lg:gap-6 mt-8">
+        { featuredGuides.map((guide) => (
```
IsaacMurtagh, line 41: In the future we will want to control the order of guides, we don't know what should control that yet so it should be fine to just have them ordered alphabetically.

ncjones: The farmlands guides should not be first

PR #694, opened 2023-03-14
Add basic search functionality

Neeraj-Ghodla:
## Description
Basic search functionality is implemented by creating an index-data.json file containing all the headings, descriptions, and anchor links at build time.

We then use the flexsearch library to build an index using the data from index-data.json to provide local search functionality.

## How to test
* Either press `/` or click on the search bar to open it
* Type a keyword to search. If the keyword exists in any heading or description, you should see some search results otherwise you'll see the "Nothing Found" text.

Desktop View

Mobile View

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -1,4 +1,5 @@
 import { defineConfig } from 'astro/config';
+import flexsearchPlugin from './src/plugins/flexsearch';
 
 // https://astro.build/config
 import tailwind from '@astrojs/tailwind';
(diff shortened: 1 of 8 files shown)


IsaacMurtagh:
Can you add a screenshot an fix pipeline?

astro.config.mjs
```diff
@@ -1,4 +1,5 @@
 import { defineConfig } from 'astro/config';
+import flexsearchPlugin from './src/plugins/flexsearch';
```
IsaacMurtagh, line 2: I thought this doesn't work, is this why the build is broken?

tom-blake:
@Neeraj-Ghodla Can you please drop some mobile/desktop screenshots and a test plan into this PR plz? 🙏

astro.config.mjs
Neeraj-Ghodla: It wasn't because of this. I've fixed the build.

IsaacMurtagh (approved):
Nice, I like it

PR #696, opened 2023-03-21
Farmlands pages adjustments

Ray2126:
These are adjustments that Farmlands has asked us to make.

### DIFF
--- a/src/content/guides/farmlands-card-partner-support.mdx
+++ b/src/content/guides/farmlands-card-partner-support.mdx
@@ -17,6 +17,7 @@ If you are a Farmlands Card Partner needing extra support this is the guide for 
 1. If you require assistance with Portal functionality, contact Centrapay at support@centrapay.com or Farmlands at card.specialist@farmlands.co.nz
 You can also use the Support window inside the Portal.
 2. If you have a query relating to an invoice or payment, please contact Farmlands on invoice.queries@farmlands.co.nz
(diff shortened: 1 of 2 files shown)


src/content/guides/farmlands-card-partner-support.mdx
```diff
@@ -17,6 +17,7 @@ If you are a Farmlands Card Partner needing extra support this is the guide for
 1. If you require assistance with Portal functionality, contact Centrapay at support@centrapay.com or Farmlands at card.specialist@farmlands.co.nz
 You can also use the Support window inside the Portal.
 2. If you have a query relating to an invoice or payment, please contact Farmlands on invoice.queries@farmlands.co.nz
+3. If you are needing to add a new branch/location please contact your Card Portfolio manager or Card.Specialist@farmlands.co.nz as Farmlands will need to complete the setup of the branch/location within our system prior to this being enabled in the Portal
```
AndyClifford, line 20: This could be reworded as `If you need to add a new...`. 

Also emails aren't case sensitive :)

AndyClifford (approved):

src/content/guides/farmlands-card-partner-support.mdx
Ray2126: This is copy-pasted from what Farmlands has asked us to put on the site. I think ultimately what goes on the site is up to them so I'm thinking I shouldn't change anything.

PR #697, opened 2023-03-21
Fix wording in integrating third party asset guide

IsaacMurtagh:
The wording around payment method, asset type, and payment option was inconsistent.

### DIFF
--- a/src/content/guides/integrating-third-party-asset.mdx
+++ b/src/content/guides/integrating-third-party-asset.mdx
@@ -8,7 +8,7 @@ nav:
 
 By integrating with Centrapay as a third-party asset provider, you can take advantage of our connections with terminals, point-of-sale systems, and merchant networks, thereby expanding the reach of your digital asset to a wider audience.
 
-Once you have defined your asset with Centrapay and implemented the required Uplink APIs, consumers will be able to spend your digital asset using one of our payment flows wherever merchants accept your digital asset.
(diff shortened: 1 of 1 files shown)


alexBinns (commented):
Should it be payment option or payment method? It seems that payment option is used more in the docs.

src/content/guides/integrating-third-party-asset.mdx
```diff
@@ -46,15 +46,15 @@ An example description is `Centrapay NZD`, `Bitcoin`, or `Ethereum`.
 
 ### Supported Currencies
 
-Payment Requests have a value determined by the currency a Merchant accepts. You should supply a finite list of three-letter [ISO currency codes](https://en.wikipedia.org/wiki/ISO_4217).
+[Payment Requests](https://docs.centrapay.com/api/payment-requests) have a value determined by the currency a Merchant accepts. You should supply a finite list of three-letter [ISO currency codes](https://en.wikipedia.org/wiki/ISO_4217) that is supported by your asset.
 
-### Example Payment Option
+### Example Defintiation
```
daallen303, line 51: I think this change may have been unintentional.

alexBinns:
> 

I guess there are certain contexts where each of them make sense one would be in a more technical aspect and the other would be in a less technical aspect

src/content/guides/integrating-third-party-asset.mdx
IsaacMurtagh: Yes, thanks :)

IsaacMurtagh: It was intentional. The H2 is called Defining a payment method, so thought it was redundant to add Payment Method to this header

laurencemcknight (approved):

PR #705, opened 2023-04-11
Add payment request fields for payment requests

subarashii-huch09:
The `barcodeType` and `collectionId` fields have been added to the create payment request endpoint as part of the requirements for implementing digital token redemption. Since we use the public API endpoint `POST /api/payment-requests`, the API documentation needs to be updated to reflect these changes.
[Notion Link ](https://www.notion.so/centrapay/Add-collectionId-barcodeType-fields-to-the-create-payment-request-endpoint-5ce079019cb049778e5de05f52d8bcea?pvs=4)

The API Design doc in Notion is also updated. Here is the [link](https://www.notion.so/centrapay/Create-Payment-Request-7bd0d884c5414823a89cdad9fed60da9?pvs=4#81ffb5f26dc04476a46d8c975c7fe658)

PR #708, opened 2023-04-13
Update payment request fields for payment requests

subarashii-huch09:
This pull request is a follow-up based on further [feedback ](https://github.com/centrapay/centrapay-docs/commit/4631bdb2a35fa3026c09854703301968713a496b) from this PR (#705) 

The following changes have been made to this pull request:

- The description of `barcodeType` has been updated.
- The value of `barcodeType` has been changed to 'ticketek'.
- An example of creating a payment request with `barcode`, `barcodeType`, and `collectionId` has been added.
- A typo has been fixed. There was an extra **|** in the optional field table, which caused the table to have an additional column that does not conform to the correct format. ( view the screenshot below)
Screenshot 2023-04-13 at 3 11 48 PM


Ray2126 (approved):
Nice work, this is looking good!

PR #713, opened 2023-04-17
Move main content in base layout out of vue component

IsaacMurtagh:
Removing app shell and instead having the main content inside the base layout. This way the main content does not need JS to load the content - should have some performance benefits as Astro recommends only wrapping components in vue if necessary to keep things fast. https://docs.astro.build/en/concepts/islands/


IsaacMurtagh:
Preview: http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/

tom-blake:
@IsaacMurtagh It doesn't look like the bug is fixed when I look at the preview. I'm also not a fan of the approach of just wrapping Vue components - it's adding complexity to the codebase without understanding the root cause of the issue 🤔 

I think our disclosure state logic may be worth looking into 👀

IsaacMurtagh:
> @IsaacMurtagh It doesn't look like the bug is fixed when I look at the preview. I'm also not a fan of the approach of just wrapping Vue components - it's adding complexity to the codebase without understanding the root cause of the issue 🤔
> 
> I think our disclosure state logic may be worth looking into 👀

Thanks, I'll look into the disclosure to fix this.

Regarding the app shell being an Astro component and wrapping the rest as a Vue component - I think this is correct.
How it is being done before this change the main content is a slot inside the app shell which requires javascript. This change would make it so only the siblings of the main content would need JS to load.

IsaacMurtagh:
> > @IsaacMurtagh It doesn't look like the bug is fixed when I look at the preview. I'm also not a fan of the approach of just wrapping Vue components - it's adding complexity to the codebase without understanding the root cause of the issue 🤔
> > I think our disclosure state logic may be worth looking into 👀
> 
> Thanks, I'll look into the disclosure to fix this.
> 
> Regarding the app shell being an Astro component and wrapping the rest as a Vue component - I think this is correct. How it is being done before this change the main content is a slot inside the app shell which requires javascript. This change would make it so only the siblings of the main content would need JS to load.

Or an alternative approach would be to remove the app shell completely and have the basic layout responsible for the main content. One less layer to think of.

PR #714, opened 2023-04-18
Add TOKEN_COLLECTION_NOT_FOUND error response

subarashii-huch09:
As part of implementing create a payment with Ticketek ticket/barcode and to enhance the handling of invalid token collections, this pull request adds a new error response named `TOKEN_COLLECTION_INVALID` to the error response table. This update aims to offer clear and concise information to users whenever this particular error arises.
<br> 
Screenshot 2023-04-18 at 2 58 25 PM

### DIFF
--- a/legacy/src/api/payment-requests/payment-requests.md
+++ b/legacy/src/api/payment-requests/payment-requests.md
@@ -496,7 +496,7 @@ Payment Activities are created when a Payment Request has been **created**, **pa
 | 400    | {% break _ CHECKSUM_FAILED %}              | Luhn checksum digit doesn't pass.                                                                         |
 | 403    | {% break _ PATRON_CODE_INVALID %}          | Patron Code doesn't exist or has expired.                                                                 |
 | 403    | {% break _ NO_AVAILABLE_PAYMENT_OPTIONS %} | No payment options match the requested payment parameters.                                                |
-| 403    | {% break _ TOKEN_COLLECTION_INVALID %}     | Either the `collectionId` is not configured with the provided token collection or the token collection does not exist.                                                                      |
+| 403    | {% break _ TOKEN_COLLECTION_NOT_FOUND %}   | The token collection does not exist.                                                                      |
 
 <a name="get-payment-request"></a>
 ### Get a Payment Request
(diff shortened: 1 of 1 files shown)


legacy/src/api/payment-requests/payment-requests.md
```diff
@@ -496,6 +496,7 @@ Payment Activities are created when a Payment Request has been **created**, **pa
 | 400    | {% break _ CHECKSUM_FAILED %}              | Luhn checksum digit doesn't pass.                                                                         |
 | 403    | {% break _ PATRON_CODE_INVALID %}          | Patron Code doesn't exist or has expired.                                                                 |
 | 403    | {% break _ NO_AVAILABLE_PAYMENT_OPTIONS %} | No payment options match the requested payment parameters.                                                |
+| 403    | {% break _ TOKEN_COLLECTION_INVALID %}     | Either the `collectionId` is not configured with the provided token collection or the token collection does not exist.                                                                      |
```
Ray2126, line 499: What does "the `collectionId` is not configured with the provided token collection" mean?

Now that I think about it, the only reason we throw this error is because the corresponding token collection does not exist. Therefore, I think the error should be `TOKEN_COLLECTION_NOT_FOUND`

Ray2126 (changes requested):

legacy/src/api/payment-requests/payment-requests.md
subarashii-huch09: I like the new error name `TOKEN_COLLECTION_NOT_FOUND`

The reason for "the `collectionId` is not configured with the provided token collection" is that there might be a scenario where the collectionId is valid but hasn't yet been set up/configured to a token collection.

 I agreed that we could implement this later and we will focus on `TOKEN_COLLECTION_NOT_FOUND` with the description of 'The token collection does not exist.' 👍

Ray2126 (approved):

Ray2126:
Let's change the title as the error message has changed :)

subarashii-huch09:
> Let's change the title as the error message has changed :)

Great spotting! Have fixed it now. Thanks :)

PR #715, opened 2023-04-18
Use current state of nav disclosure as source of open state

IsaacMurtagh:
This fixes the bug where the site nav disclosure was not opening when pressing it for the first time on load.
This occurred because the state of the nav was open when reading from the local storage however it was not reflecting this visually as the calculation occurred before the component was mounted if read from local storage.

This change fixes this by reading the local storage inside the onMounted lifecycle hook.

### DIFF
--- a/src/components/SiteNavigationDisclosure.vue
+++ b/src/components/SiteNavigationDisclosure.vue
@@ -73,7 +73,7 @@
 </template>
 
 <script setup>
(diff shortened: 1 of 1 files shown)


IsaacMurtagh:
Preview: http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/

laurencemcknight (approved):

PR #716, opened 2023-04-18
Define order of side bar headings

laurencemcknight:
Matthias has requested that "Centrapay Experiences" should be the first heading under "Reference" in the side bar.
This change allows side bar headings to be re-ordered, so that "Centrapay Experiences" can be moved to the top.

Testing:
Assert that "Centrapay Experiences" is the first heading under "Reference" in the side bar.
Assert that all other headings still appear and perform correctly when clicked.

### DIFF
--- a/src/utils/Navigation.js
+++ b/src/utils/Navigation.js
@@ -13,13 +13,20 @@ class Navigation {
           title: 'Reference',
           to: '/reference',
           icon: 'Receipt',
-          children: [],
+          children: [
+            { title: 'Centrapay Experiences', children: [] },
+            { title: 'Digital Assets', children: [] },
+            { title: 'App Integrations', children: [] },
+            { title: 'Merchant Integrations', children: [] },
+          ],
         },
         {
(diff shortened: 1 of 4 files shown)


src/utils/Navigation.js
```diff
@@ -13,13 +13,20 @@ class Navigation {
...
+            { title: 'Merchant Integrations', children: [] },
+          ],
         },
         {
           title: 'Connections',
           to: '/connections',
           icon: 'Connections',
-          children: [],
+          children: [
+            { title: 'Farmlands', to: '/connections/farmlands', children: [] },
```
laurencemcknight, line 28: A consequence of this change is that if we want clicking on a heading to route to a page that is not the first page in the heading, we have to explicitly define that page's route here.
I don't think this is an issue as we are heading towards defining the navigation structure more explicitly anyway.

daallen303: Isn't that already the behavior that exists?

PR #717, opened 2023-04-21
Define order of side nav disclosures

laurencemcknight:
We need the ability to order disclosures in the navigation side bar. This PR allows us to do that.

After talking to designers and Raymond we concluded that the current code for the nav side bar was unnecessarily complex. We therefore decided to refactor the side bar while we were in the area.

Testing:
Assert that the side bar can be interacted with as normal.
Assert that side bar components that are currently selected have a grey background.
Assert that clicking the API link in the sidebar routes to the API web page.


IsaacMurtagh (approved):

laurencemcknight:
Preview: http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com

PR #722, opened 2023-05-14
Guide for UaaS JWT authorisation 

laurencemcknight:
image

### DIFF
--- a/src/content/guides/integrating-third-party-asset.mdx
+++ b/src/content/guides/integrating-third-party-asset.mdx
@@ -130,46 +130,30 @@ To verify a request start by decoding the JWT.
 
 Use the decoded JWT fields to validate the following:
 
-Expiry
+Exp is provided as a default expiry. Alternatively, use iat + your own expiry window to determine if the JWT has expired.
 
-Verify that the request authorisation has not expired to prevent against replay attack.
+Assert that the audience is correct by checking that aud is equal to the base URL of your Uplink API.
 
-Exp is provided as a default expiry, with a 5 minute expiry window. Check that the default expiry has not passed by asserting that exp is greater than the current Unix timestamp.
-
-To use a custom expiry window instead of the default provided in exp, add the duration of the expiry window in seconds to iat. To verify this custom expiry has not passed, assert that the value is greater than the current Unix timestamp.
-
-
-Audience
-
-To prevent against man-in-the-middle attacks assert that aud is equal to the base URL of your Uplink API.
-
-
-Request payload
-
-The request_body_sha256 property can be used to verify that the request payload has not been tampered with. Request_body_sha256 contains a hash of the request payload, created using the [SHA256 algorythm](https://en.wikipedia.org/wiki/SHA-2).
-We can therefore assert the payload is unchanged by hashing the received request payload using the SHA256 algorythm and checking for equality with request_body_sha256.
-
-**Example Of Verifying A Request Payload**
+The request_body_sha256 property should be used to verify that the request payload has not been tampered with. This should be done by hashing the received request payload using the SHA256 algorythm and checking for equality with request_body_sha256 as shown below.
 
 ```javascript
-  const crypto = require('crypto');
+  const crypto = require("crypto");
 
   const decodedJwt = {
-    request_body_sha256: 'd4735e3a265e16eee03f59718b9b5d03019c07d8b6c51f90da3a666eec13ab35',
+    request_body_sha256: "d4735e3a265e16eee03f59718b9b5d03019c07d8b6c51f90da3a666eec13ab35",
   };
 
   const payload = {
-    vader: 'I am your father',
-    luke: 'That\'s not true, that\'s impossible',
+    vader: "I am your father",
+    luke: "That's not true, that's impossible",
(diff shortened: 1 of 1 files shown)


src/content/guides/integrating-third-party-asset.mdx
```diff
@@ -108,6 +108,29 @@ curl https://auth.centrapay.com/api/.well-known/jwks.json
 }
 ```
 
+To verify a request start by decripting the JWT.
```
tom-blake, line 111: @laurencemcknight Typo here - "decripting"

src/content/guides/integrating-third-party-asset.mdx
```diff
@@ -108,6 +108,29 @@ curl https://auth.centrapay.com/api/.well-known/jwks.json
...
+
+**Decripted JWT Payload**
```
tom-blake, line 113: @laurencemcknight There is a sentence above about verifying the JWT using the public key. I think this needs restructuring - worth looking at how other companies document this (e.g. [Stripe](https://stripe.com/docs/webhooks/signatures))

src/content/guides/integrating-third-party-asset.mdx
```diff
@@ -108,6 +108,29 @@ curl https://auth.centrapay.com/api/.well-known/jwks.json
...
+  "aud": "https://your.endpoint/",
+  "request_body_sha256": "b9195bf41bf0e38ab0ab44e7ef5b9af5cb0fe2ece8dee5d112d7485bf4ef0007",
+}
+```
+
+Use the decripted JWT validate the following fields:
+
+|        Field        |                                                                         Assertion                                                                           |
+| ------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| exp                 | Assert that exp is greater than the current Unix timestamp.                                                                                                 |
```
tom-blake, line 128: @laurencemcknight I thought we were including the `iat` field too?

We should document what our expiry is (e.g. 5 mins from the time that the JWT was created) + provide the `iat` field to allow integrators to use a shorter expiry if they wish

src/content/guides/integrating-third-party-asset.mdx
```diff
@@ -108,6 +108,29 @@ curl https://auth.centrapay.com/api/.well-known/jwks.json
...
+| aud                 | Assert that aud is equal to the base URL of your Uplink API.                                                                                                |
+| request_body_sha256 | Hash the request payload using the [SHA256 algorythm](https://en.wikipedia.org/wiki/SHA-2). Assert that request_body_sha256 is equal to the hashed payload. |
```
tom-blake, line 130: @laurencemcknight We should be really explicit with should integrators must do - verifying the payload is important

src/content/guides/integrating-third-party-asset.mdx
IsaacMurtagh: Decrypting is not the right word. Decoding is more accurate

src/content/guides/integrating-third-party-asset.mdx
IsaacMurtagh: @laurencemcknight You could give an example how to do it in JS as it's important they know they need to stringify the json

src/content/guides/integrating-third-party-asset.mdx
```diff
@@ -108,6 +108,29 @@ curl https://auth.centrapay.com/api/.well-known/jwks.json
...
 
+To verify a request start by decripting the JWT.
+
+**Decripted JWT Payload**
+
+```JSON
+{
+  "iat": "1684105185",
+  "exp": "1684105485",
+  "aud": "https://your.endpoint/",
```
IsaacMurtagh, line 119: Can you confirm we will be sending a trailing slash?

src/content/guides/integrating-third-party-asset.mdx
```diff
@@ -108,6 +108,29 @@ curl https://auth.centrapay.com/api/.well-known/jwks.json
...
+  "request_body_sha256": "b9195bf41bf0e38ab0ab44e7ef5b9af5cb0fe2ece8dee5d112d7485bf4ef0007",
+}
+```
+
+Use the decripted JWT validate the following fields:
+
+|        Field        |                                                                         Assertion                                                                           |
```
IsaacMurtagh, line 126: I think this table is better used for explaining the value and then a paragraph on each of the fields about how to make the assertion.

IsaacMurtagh (approved):

src/content/guides/integrating-third-party-asset.mdx
daallen303: Hmm, I think it's more decrypting. We encrypt the message with a specific private key and then they need the public key in order to decrypt the message right?

daallen303: I think encoding/decoding seems more like applying an algorithm without any private/public keys.

laurencemcknight: @daallen303 "Encoding" and "decoding" is the common terminology online used for JWTs. Additionally we already use that terminology in our docs for JWTs.

PR #726, opened 2023-05-16
Add collections to token payment option for paymentRequest response

alexBinns:
This PR adds the collections to the payment request response to allow payment with tokens against a collection

### DIFF
--- a/legacy/src/api/payment-requests/payment-requests.md
+++ b/legacy/src/api/payment-requests/payment-requests.md
@@ -103,13 +103,13 @@ version (documented on this page) and the "legacy" version (documented at
 
 {% h4 Optional Fields %}
 
-|     Field      |  Type  |                                        Description                                         |
-| -------------- | ------ | ------------------------------------------------------------------------------------------ |
-| bitcoinAddress | String | ★  Address to send Bitcoin, when the "assetType" is `bitcoin.*`                            |
-| cennzAddress   | String | ★  Address to send CPay, when the "assetType" is `cennznet.*`                              |
-| wavesAddress   | String | ★  Waves address to send Zap tokens, when the "assetType" is `zap.*`                       |
-| productCodes   | Array  | Supported product codes for the payment request, when the "assetType" is `epay.nzd.*`      |
-| collectionIds    | Array  | Supported collectionIds for the payment request, when the "assetType" is `centrapay.token.*` |
+|     Field      |  Type  |                                         Description                                          |
+| -------------- | ------ | -------------------------------------------------------------------------------------------- |
+| bitcoinAddress | String | ★  Address to send Bitcoin, when the "assetType" is `bitcoin.*`                              |
+| cennzAddress   | String | ★  Address to send CPay, when the "assetType" is `cennznet.*`                                |
+| wavesAddress   | String | ★  Waves address to send Zap tokens, when the "assetType" is `zap.*`                         |
+| productCodes   | Array  | Supported product codes for the payment request, when the "assetType" is `epay.nzd.*`        |
+| collectionIds  | Array  | Supported collectionIds for the payment request, when the "assetType" is `centrapay.token.*` |
 
 ★  For payment options which specify an address, there's a requirement to make a transaction on an external ledger.
 Once you have made that payment, you can use the transaction id to [Pay a Payment Request](#pay) using the legacy payment API.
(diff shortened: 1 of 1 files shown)


legacy/src/api/payment-requests/payment-requests.md
```diff
@@ -458,6 +459,15 @@ Payment Activities are created when a Payment Request has been **created**, **pa
       "amount": "6190",
       "assetType": "epay.nzd.test",
       "productCodes": [ "23403" ]
+    },
+    {
+      "amount": "6190",
+      "assetType": "centrapay.token.test",
+      "collections: [
+        {
+          "id": "345224"
```
laurencemcknight, line 468: Do we need to return the collection name or image so that may be displayed in the front end?

alexBinns: I don't think so as we don't do it for epay

alexBinns: The brand img etc is more for displaying in the app

legacy/src/api/payment-requests/payment-requests.md
```diff
@@ -103,12 +103,13 @@ version (documented on this page) and the "legacy" version (documented at
...
-| wavesAddress   | String | ★  Waves address to send Zap tokens, when the "assetType" is `zap.*`                  |
-| productCodes   | Array  | Supported product codes for the payment request, when the "assetType" is `epay.nzd.*` |
+|     Field      |  Type  |                                        Description                                         |
+| -------------- | ------ | ------------------------------------------------------------------------------------------ |
+| bitcoinAddress | String | ★  Address to send Bitcoin, when the "assetType" is `bitcoin.*`                            |
+| cennzAddress   | String | ★  Address to send CPay, when the "assetType" is `cennznet.*`                              |
+| wavesAddress   | String | ★  Waves address to send Zap tokens, when the "assetType" is `zap.*`                       |
+| productCodes   | Array  | Supported product codes for the payment request, when the "assetType" is `epay.nzd.*`      |
+| collections    | Array  | Supported collections for the payment request, when the "assetType" is `centrapay.token.*` |
```
Ray2126, line 112: Why is this an array of objects rather than an array of `collectionId`s?

legacy/src/api/payment-requests/payment-requests.md
```diff
@@ -529,21 +539,30 @@ Payment Activities are created when a Payment Request has been **created**, **pa
...
+      "discount": "199",
+    },
+    {
+      "amount": "6190",
+      "assetType": "centrapay.token.test",
+      "collections": [
+        {
+          "id": "345224"
+        }
+      ]
```
Ray2126, line 564: Is this in the wrong place? I would expect this to be in the `paymentOptions` array.

Ray2126 (changes requested):

legacy/src/api/payment-requests/payment-requests.md
alexBinns: Yup

legacy/src/api/payment-requests/payment-requests.md
laurencemcknight: I agree with @Ray2126, this should be an array of collectionIds. I also think that the field should be renamed to collectionIds.

Ray2126 (approved):

legacy/src/api/payment-requests/payment-requests.md
```diff
@@ -103,12 +103,13 @@ version (documented on this page) and the "legacy" version (documented at
...
-| wavesAddress   | String | ★  Waves address to send Zap tokens, when the "assetType" is `zap.*`                  |
-| productCodes   | Array  | Supported product codes for the payment request, when the "assetType" is `epay.nzd.*` |
+|     Field      |  Type  |                                        Description                                         |
+| -------------- | ------ | ------------------------------------------------------------------------------------------ |
+| bitcoinAddress | String | ★  Address to send Bitcoin, when the "assetType" is `bitcoin.*`                            |
+| cennzAddress   | String | ★  Address to send CPay, when the "assetType" is `cennznet.*`                              |
+| wavesAddress   | String | ★  Waves address to send Zap tokens, when the "assetType" is `zap.*`                       |
+| productCodes   | Array  | Supported product codes for the payment request, when the "assetType" is `epay.nzd.*`      |
+| collectionIds    | Array  | Supported collectionIds for the payment request, when the "assetType" is `centrapay.token.*` |
```
IsaacMurtagh, line 112: formatting needs fixing

IsaacMurtagh (approved):

PR #734, opened 2023-06-08
Add partial pay example to create payment requests

alexBinns:
It was found that we didn't have an example for partial pay. This adds that example curl.

### DIFF
--- a/legacy/src/api/payment-requests/payment-requests.md
+++ b/legacy/src/api/payment-requests/payment-requests.md
@@ -371,6 +371,42 @@ Payment Activities are created when a Payment Request has been **created**, **pa
     })
   }
   example {
+    title 'Create a Payment Request with Line Items and Partial Pay'
+    body ({
+      configId: '5efbe2fb96c08357bb2b9242',
+      value: { amount: '6190', currency: 'NZD' },
+      partialAllowed: true,
+      basketAmount: '10000',
+      lineItems: [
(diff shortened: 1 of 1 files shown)


legacy/src/api/payment-requests/payment-requests.md
```diff
@@ -787,6 +823,15 @@ Use authorization to authorize an external transaction.
       "authorization": "LiOkwEtDGTVz4ESv5NS83",
     })
   }
+  example {
+    title 'Partially Pay a Payment Request'
+    body ({
+      "assetType": "centrapay.nzd.main",
+      "assetId": "WRhAxxWpTKb5U7pXyxQjjY",
+      "amount": "200",
+      "mode": "partial"
```
IsaacMurtagh, line 832: This is actually partial-payment

IsaacMurtagh (commented):
We have a problem where we're not specifying what properties are optional/required when calling payment requests APIs, I think we should do a follow up PR which would add this info for the APIs.

Example: https://docs.centrapay.com/api/bank-accounts#creating-a-bank-account

IsaacMurtagh (approved):

legacy/src/api/payment-requests/payment-requests.md
alexBinns: Good catch fixed

PR #755, opened 2023-07-11
Add merchantConfigId to refund API

laurencemcknight:
merchantConfigId is a new optional field we are adding to the refund API. This will allow Farmlands to associate refunds with specific merchants, instead of defaulting to the merchant used to create the payment.


GeekOnGadgets (approved):
LGTM

PR #759, opened 2023-07-21
Add error for invalid account type on update subscription route

Ray2126:


Ray2126:
The `quartz` subscription is applied to individual account.

PR #761, opened 2023-07-28
Document refund payload fields

laurencemcknight:
I noticed our refund API docs were missing documentation of payload fields (which is present for all other APIs). This PR adds them.

image


legacy/src/api/payment-requests/payment-requests.md
```diff
@@ -1014,6 +1014,7 @@ One of either `assetId`, `transactionId` or `authorization` must be provided.
         "amount": "6190",
         "currency": "NZD",
       },
+      "externalRef": "e8df06e2-13a5-48b4-b670-3fd6d815fe0a",
```
laurencemcknight, line 1017: This field must be added to the example as it is a required field for the API.

alexBinns (approved):

PR #766, opened 2023-08-29
Add note to not call release when there is nothing to release

alexBinns:


src/content/guides/farmlands-pos-integration.mdx
```diff
@@ -722,6 +722,10 @@ Note over POS: ✅ Display Release confirmation
     ```
 
 6. The POS [releases any remaining authorised funds](https://docs.centrapay.com/api/payment-requests#release-funds-held-for-a-pre-auth-payment-request-experimental) against the [Payment Request](https://docs.centrapay.com/api/payment-requests#payment-request) back to the Cardholder.
+Funds requiring release can either be tracked internally or you can [list payment activities](https://docs.centrapay.com/api/payment-requests#list-payment-activities-for-a-payment-request) and sum the confirmation [payment activities](https://docs.centrapay.com/api/payment-requests#payment-activity).
```
laurencemcknight, line 725: IMO this line is self explanatory and does not need to be documented.

src/content/guides/farmlands-pos-integration.mdx
```diff
@@ -722,6 +722,10 @@ Note over POS: ✅ Display Release confirmation
...
+
+> Releasing funds should only happen if there are remaining funds to release.
```
laurencemcknight, line 727: Isn't this already explained by "releases any remaining authorised funds" (line 724).

src/content/guides/farmlands-pos-integration.mdx
alexBinns: This was requested by Brijesh to make it easier for an integrator to know how to know the remaining left on a pre auth. He wanted a whole section on it but I think this is enough. I don't think that it would be immediatly obvious how to get the remaining amount on a pre auth. Perhaps the part about tracking internally could be removed.

src/content/guides/farmlands-pos-integration.mdx
alexBinns: This is a request by Farmlands as there have been multiple integrators that have wanted to do this.

laurencemcknight: Ok that makes sense then

src/content/guides/farmlands-pos-integration.mdx
laurencemcknight: Yes, I would remove the part about internal tracking.

laurencemcknight (approved):

PR #767, opened 2023-09-04
docs: update merchant location description

Neeraj-Ghodla:
## Description

Updated the description of the location field on the merchant model recommending users to provide the location of the merchant when creating them, otherwise risk not appearing in the results when users search for merchants.

| Before | After |
|--------|--------|
| Screen Shot 2023-09-04 at 15 41 10 | Screen Shot 2023-09-04 at 15 40 47 |


mark-scott:
@Neeraj-Ghodla I'll wait til this is amended based on the discussion about using address as the source of truth

mark-scott (approved):

PR #782, opened 2023-10-02
Draft page

AndyClifford:
Created a wrapper function for getCollection which does not return any draft pages when in production. This allows us to incrementally migrate the legacy API reference.

**Ticket:** https://www.notion.so/centrapay/Create-a-draft-Astro-page-b39e82f4a15a41ef8b34c41fa45d54d3?pvs=4

**Test Plan:**
- [ ] Navigate to docs.centrapay.com/guides/draft - This should route you to the legacy Jekyll 404 page
- [ ] Assert that the http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/guides/draft is visible and does not redirect to 404 page

**The draft page shows up in DEV mode**
image

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -26,7 +26,7 @@ jobs:
         with:
           node-version: 18
       - run: yarn install --frozen-lock-file
-      - run: ./build.sh http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com
+      - run: ./build.sh
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v2
         with:
--- a/build.sh
+++ b/build.sh
@@ -2,7 +2,17 @@
 
 set -euo pipefail
 
-url="${1:-}"
+mode="development"
+
+while [ $# -gt 0 ]; do
+  case $1 in
+    --prod)
+      shift
+      mode="production"
+    ;;
+  esac
+  shift
+done
 
 cd legacy
 
@@ -11,8 +21,8 @@ bundle exec jekyll build
 
 cd ../
 
-if [[ -n "$url" ]]; then
-  yarn build --site "$url"
+if [ "$mode" == "development" ]; then
+  yarn build --mode "$mode" --site "http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com"
 else
   yarn build
 fi
--- a/src/content/api/introduction.mdx
+++ b/src/content/api/introduction.mdx
@@ -1,0 +1,20 @@
+---
+title: API Introduction
+description: Introduction to the API Reference
+nav:
+  path: API/Getting Started
+  order: 1
+---
+
+The Centrapay API is an [RMM](https://en.wikipedia.org/wiki/Richardson_Maturity_Model)
+level 2 RESTful web service which expresses operations in terms of HTTP verbs on
+resource-oriented URLs. API endpoint definitions in these docs are grouped by resource
+type along with definitions for the associated resource types.
+
+Most API calls require [authentication](https://docs.centrapay.com/api/auth) using an API key or
+JWT. HTTP requests and responses usually have JSON payloads and use
+"application/json" as the content type.
+
+Some API features may be flagged as **EXPERIMENTAL**. These API features may be
+removed or changed without warning and should not be relied on in a production
+setting.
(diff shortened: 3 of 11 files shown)


src/content/guides/draft.mdx
MeganSteenkamp, line 1: ❓ Question: @AndyClifford what happens if this draft page ends up with the same URL as an existing Jekyll page?

AndyClifford: Looking into this now

AndyClifford: The dev site does not render the jekkyl site. If this is expected behaviour then this won't be an issue. Will confirm this

.github/workflows/deploy-dev.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         with:
           node-version: 18
       - run: yarn install --frozen-lock-file
-      - run: ./build.sh http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com
+      - run: ./build.sh http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com development
```
MeganSteenkamp, line 29: There are other workflows where we run `build.sh` like `validate.yml` where we will want to validate links. I'm wondering if this should default to development and then override for production. Do you think defaulting to `production` is safer because we won't accidentally show draft pages? 🤔

src/content/config.ts
```diff
@@ -16,4 +17,5 @@ const guidesCollection = defineCollection({
 export const collections = {
   'guides': guidesCollection,
   'connections': guidesCollection,
+  'api': guidesCollection
```
MeganSteenkamp, line 20: ✏️ Suggestion: Maybe we should rename this variable to something more generic like `collection`.

src/content/guides/draft.mdx
MeganSteenkamp: @AndyClifford do we still need this page now that you have a draft API page?

AndyClifford: Good call I'll remove this

.github/workflows/deploy-dev.yml
AndyClifford: Do we want to validate links of draft pages?

MeganSteenkamp: I would think so - you'll want to know when you first make the page whether you've put in an invalid link.

tom-blake, line 29: @AndyClifford Suggestion: Only pass `mode` to the script. Let the script handle the extra complexity of setting the `--site` based on the mode

tom-blake: [Vite defaults to development mode](https://vitejs.dev/guide/env-and-mode.html#modes)

Validating links in draft pages gives us a faster feedback loop

src/content/api/introduction.mdx
```diff
@@ -0,0 +1,9 @@
+---
+title: Introduction - Draft
+description: Draft Introduction
+nav:
+  path: API/Getting Started
+  order: 1
+---
+
+Setting `draft: true` in the front matter of the guide means this page will not be published in production. This allows us to incrementally migrate the legacy API reference. We will remove this once we migrate the first api reference page.
```
tom-blake, line 9: @AndyClifford Suggestion: Use the content in https://docs.centrapay.com/api/introduction

src/utils/getCollection.js
```diff
@@ -0,0 +1,7 @@
+import { getCollection as get } from 'astro:content';
+
+export async function getCollection(collection) {
+  return await get(collection, ({ data }) => {
+    return import.meta.env.MODE === 'production' ? data.draft !== true : true;
```
tom-blake, line 5: @AndyClifford This can be simplified

Change: `return import.meta.env.MODE !== 'production' || !data.draft`

build.sh
```diff
@@ -2,7 +2,17 @@
 
 set -euo pipefail
 
-url="${1:-}"
+mode="development"
```
AndyClifford, line 5: @MeganSteenkamp I have defaulted to development here - what do you think?

AndyClifford: I prefer not to override their default of `production` mode but happy either way :)

build.sh
```diff
@@ -2,7 +2,17 @@
...
+
+while [ $# -gt 0 ]; do
+  case $1 in
+    --prod)
+      shift
+      mode="production"
```
MeganSteenkamp, line 11: @AndyClifford do we need this if production is default?

build.sh
MeganSteenkamp: @AndyClifford I think this is ok 👍

MeganSteenkamp (approved):

PR #783, opened 2023-10-02
⛓️ Check build for broken links

MeganSteenkamp:
- Fail build when Astro site contains broken links.
- Add ability to test for broken static links.

[📖 Detect Broken Links](https://www.notion.so/centrapay/Detect-broken-links-9dc17ee52456481e97db0e1847d902d8)

🧪 Test plan: Expect to see build step passing in pipeline.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,4 +1,4 @@
-name: Validate Jekyll Site
+name: Validate Site
 
 on:
   workflow_dispatch:
@@ -7,10 +7,10 @@ on:
       - master
 
 jobs:
-  jekyll:
+  validate:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
@@ -29,3 +29,4 @@ jobs:
       - run: yarn lint
       - run: yarn test
       - run: ./build.sh
+      - run: yarn linkinator
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "private": true,
   "scripts": {
     "lint": "eslint --ext .js,.mjs,.ts,.vue .",
+    "linkinator": "linkinator dist/ --recurse --verbosity error --skip 'twitter.com|www.farmlands.co.nz|www.instagram.com|yourapp.example.com|centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/api'",
     "dev": "astro dev",
     "build": "astro build",
     "preview": "astro preview",
@@ -28,6 +29,7 @@
     "flexsearch": "^0.7.31",
     "glob": "^8.1.0",
     "gray-matter": "^4.0.3",
+    "linkinator": "^5.0.2",
     "mdast-util-to-string": "^3.1.1",
     "mermaid": "^9.3.0",
     "remark-html": "^15.0.2",
--- a/src/pages/api/[...slug].astro
+++ b/src/pages/api/[...slug].astro
@@ -15,9 +15,10 @@ export async function getStaticPaths() {
 
 const { entry } = Astro.props;
 const { Content, headings } = await entry.render();
+const previewImg = entry.data.img || '/default-cover.jpg';
 const { title, description } = entry.data;
 ---
-<BaseLayout title={title} description={description}>
+<BaseLayout title={title} description={description} img={previewImg}>
   <div class="relative mx-auto desktop-gutters flex justify-center">
     <div class="min-w-0 max-w-2xl flex-auto px-8 pb-16 pt-8 xl:pt-16 lg:max-w-none">
       <article>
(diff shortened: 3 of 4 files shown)


linkinator.config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "path": "/dist",
+  "recurse": true,
+  "skip": [
```
MeganSteenkamp, line 4: There are some external links that fail, I'm not sure why they return non-success status codes 🤔 
This happens both locally and in the pipeline. It also happens when you use link checkers on our live site i.e.
```
npx linkinator https://docs.centrapay.com/
```

package.json
```diff
@@ -28,6 +29,7 @@
     "flexsearch": "^0.7.31",
     "glob": "^8.1.0",
     "gray-matter": "^4.0.3",
+    "linkinator": "^5.0.2",
```
MeganSteenkamp, line 32: I chose [linkinator](https://www.npmjs.com/package/linkinator) over [hyperlink](https://www.npmjs.com/package/hyperlink) because:
- It's actively maintained (1 month last publish vs 2 years)
- It's more popular (76,174 vs 16,758 weekly downloads)
- The documentation is better IMO
- The output when running locally is also way nicer 🔥

linkinator.config.json
MeganSteenkamp: Not sure if this is okay to leave or if it's a sign these links are wrong 👀

.github/workflows/validate.yml
```diff
@@ -29,3 +29,7 @@ jobs:
       - run: yarn lint
       - run: yarn test
       - run: ./build.sh
+      - uses: JustinBeckwith/linkinator-action@v1
+        with:
+          paths: 'dist'
```
MeganSteenkamp, line 34: Only checking the Astro site here. I think this is ok as the Jekyll site will be removed soon 💅

.github/workflows/validate.yml
```diff
@@ -29,3 +29,7 @@ jobs:
       - run: yarn lint
       - run: yarn test
       - run: ./build.sh
+      - uses: JustinBeckwith/linkinator-action@v1
```
tom-blake, line 32: @MeganSteenkamp Suggestion - use a script instead of relying on this action being maintained

tom-blake: Perhaps the config could also be encapsulated in the script to avoid adding another config file to the root of the repo 👀

package.json
MeganSteenkamp: Will follow up with an ADR on this 👂

.github/workflows/validate.yml
MeganSteenkamp: Thanks @tom-blake I have now addressed this 🤠

.github/workflows/validate.yml
```diff
@@ -7,10 +7,10 @@ on:
       - master
 
 jobs:
-  jekyll:
+  validate:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
```
MeganSteenkamp, line 13: This action was generating a warning because it uses node 12 - have upgraded 🧹

package.json
```diff
@@ -2,6 +2,7 @@
   "private": true,
   "scripts": {
     "lint": "eslint --ext .js,.mjs,.ts,.vue .",
+    "linkinator": "linkinator dist/ --recurse --skip 'twitter.com|www.farmlands.co.nz|www.instagram.com'",
```
MeganSteenkamp, line 5: https://github.com/centrapay/centrapay-docs/pull/783#discussion_r1342221022

MeganSteenkamp: I have manually checked the links, check they match what Google search does so not sure what else we could do here.

tom-blake: @MeganSteenkamp It'd be good to validate external links 👀 

If linkinator can't support this, it is worth considering alternatives

tom-blake, line 5: @MeganSteenkamp Change: `--skip '$https://'

Avoids devs needing to update this config each time a new external link is added

AndyClifford: @MeganSteenkamp what does recurse do? Is it possible that it is recursing on the farmlands page? Also would using a configuration file be useful? Looks like you can specify most options including `skip` https://www.npmjs.com/package/linkinator

AndyClifford: @MeganSteenkamp what does recurse do? Is it possible that it is recursing on the external pages? Also would using a configuration file be useful? Looks like you can specify most options including `skip` https://www.npmjs.com/package/linkinator

AndyClifford: @MeganSteenkamp looks like you can specify verbosity to `debug` which might help you if you haven't done this already

MeganSteenkamp: @tom-blake I've tried other tools and they all reproduce errors for the same links so I don't think it's to do with the tooling. 
Farmlands website ➡️ 0 (HTTP_UNDEFINED)
Twitter ➡️ 403 
Instagram ➡️ Failed only in the pipeline, I'll retry this.


@AndyClifford 
* Recurse performs recursion in the `dist` directory so it will go into the guides folder, and folders beneath that etc. 
* Yes I looked at debug option and it shows HTTP responses but didn't help do any additional debugging. 
* I had a config file but had feedback from @tom-blake that it would be nicer as a script to get rid of another config file 🧹

MeganSteenkamp: @tom-blake there are some external links that are working i.e. LinkedIn 🤔

package.json
```diff
@@ -2,6 +2,7 @@
   "private": true,
   "scripts": {
     "lint": "eslint --ext .js,.mjs,.ts,.vue .",
+    "linkinator": "linkinator dist/ --recurse --verbosity error --skip 'twitter.com|www.farmlands.co.nz|www.instagram.com|yourapp.example.com|centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/api'",
```
MeganSteenkamp, line 5: The API link can be removed when the migration is finished 🤨

AndyClifford (approved):

PR #793, opened 2023-10-05
Detect if new changes break existing links

AndyClifford:
Added a unit test which determines if any static links have been broken by iterating over the build output and comparing it to a snapshot of existing links. The test will fail if a page is removed and no redirect is added for it.

This allows us to migrate the legacy docs site to the new astro site with confidence. Also ensures we add the existing legacy redirects to the new astro redirects once we publish the migration

**Test Plan:**
- [ ] CI Pipeline passes


AndyClifford:
@MeganSteenkamp @tom-blake just been playing around with this - Do you think this is necessary? Means developers have to run `yarn build --mode development` then `yarn test:build -u` every time they add a new page / redirect

src/__tests__/__snapshots__/links.test.js.snap
```diff
@@ -0,0 +1,39 @@
+// Vitest Snapshot v1
+
+exports[`the build should not break any links 1`] = `
+Set {
```
MeganSteenkamp, line 4: There must be more links than this 🤔
What about all the headings? I know there are some experimental links too.

MeganSteenkamp:
> @MeganSteenkamp @tom-blake just been playing around with this - do you think it's necessary? I think it should only run on the production build but this is difficult as the `validate` workflow uses a development build

It would be nice to have to get some extra confidence as long as it doesn't take too much time to make :) Perhaps running on the development build is ok?

src/__tests__/__snapshots__/links.test.js.snap
AndyClifford: I initially included functionality to detect broken anchor tags but removed it to reduce complexity - I didn't think breaking an anchor tag is as much of an issue. Do redirects work for anchor tags? If we expect developers to add redirects for anchor tags then I can add this back in 😀

MeganSteenkamp: Yeah I guess as long as you're on the right page it doesn't matter 😄  We are still missing lots of /api links though from the legacy site

AndyClifford: Do you think we need this test to have confidence in the migration? If so can traverse the legacy site too

MeganSteenkamp: @AndyClifford I think that would be the main thing we would want it for 👍

tom-blake: @AndyClifford Suggestion: Confirm if redirects work for anchor tags with Astro. 

Agree that broken anchors are less of a problem than broken links but they still contribute to a lesser experience. I'd prefer they were included in any checks ❤️

.github/workflows/validate.yml
```diff
@@ -29,3 +29,4 @@ jobs:
       - run: yarn lint
       - run: yarn test
       - run: ./build.sh
+      - run: yarn test:build
```
tom-blake, line 32: @AndyClifford Suggestion: Call these "integration" tests. "Build" is an overloaded term and it's not clear what it means in this context

src/__tests__/__snapshots__/links.test.js.snap
AndyClifford: I have added support for anchor tags

AndyClifford: Redirects do not support anchor tags but this is okay - we can validate changes in the snapshot to ensure redirects have been added properly

src/__tests__/links.integration.test.js
```diff
@@ -0,0 +1,40 @@
...
+  const fileContents = fs.readFileSync(filePath, 'utf-8');
+  const dom = new JSDOM(fileContents);
+  const links = Array.from(dom.window.document.querySelectorAll('a'))
+    .map(link => link.href)
+    .filter(href => href && href.startsWith('about:blank#') && href !== 'about:blank#')
+    .map(href => `${urlPath}${href.split('about:blank')[1]}`);
+  return links;
+}
+
+function extractLinksRecursively(directory) {
```
tom-blake, line 16: @AndyClifford Change: These functions should be utilities and have their own tests

src/__tests__/links.integration.test.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable complexity */
+const { extractLinksRecursively } = require('../utils/extractLinksRecursively');
+
+test('the build should not break any links', () => {
+  const uniqueLinks = extractLinksRecursively('_site');
```
MeganSteenkamp, line 5: @AndyClifford does this contain all the astro and jekyll links?

AndyClifford: Yep all of the unique links and anchor tags on each page

src/utils/extractLinksRecursively.js
```diff
@@ -0,0 +1,38 @@
...
+    } else if (stat.isFile() && filePath.endsWith('.html')) {
+      const urlPath = filePath.substring(filePath.indexOf('/'), filePath.lastIndexOf('.'));
+      links = new Set([...links, urlPath, ...extractAnchors(filePath, urlPath)]);
+    }
+  }
+  return links;
+}
+
+module.exports = {
+  extractLinksRecursively
```
tom-blake, line 37: @AndyClifford Suggestion: Rename this to `ListSiteLinks` - "recursive" is an implementation detail that consumers don't care about

src/utils/__tests__/extractLinksRecursively.test.js
```diff
@@ -0,0 +1,15 @@
...
+describe('extractLinksRecursively', () => {
+  it('should return a unique set of links and anchor tags', () => {
+    expect(extractLinksRecursively('src/utils/__tests__/examples/duplicate-paths')).toMatchSnapshot();
+  });
+
+  it('should not include any external links', () => {
+    expect(extractLinksRecursively('src/utils/__tests__/examples/external-links')).toMatchSnapshot();
+  });
+
+  it('should not include anchor tags that do not go anywhere', () => {
```
tom-blake, line 12: @AndyClifford I consider these links to be broken (either pointless or they could be a mistake). Probably another good requirement for a broken link detection tool

Suggestion: Fix these in a follow up if any are coming from the Astro site. Remove this test case when the Jekyll site is deleted

tom-blake (approved):

src/utils/__tests__/extractLinksRecursively.test.js
AndyClifford: Yeah looks like the Jekyll build automatically generates these random anchor tags e.g. `<a href="#" class="nav-list-expander">`. This isn't a problem with the astro site so can remove it after the migration

PR #796, opened 2023-10-05
404 Page

AndyClifford:
404 Page as per the Figma design https://www.figma.com/file/IgRlQWJTyGoZhlF4fVKD6B/Centrapay-Docs-(Tailwind)?type=design&node-id=636-5747&mode=design&t=a1A4fYhOc8PuRO7Y-0

Reworded the text to match the [Tailwind UI Protocol](https://protocol.tailwindui.com/fedd)

**Test Plan:**
- [ ] Go to docs.centrapay.com/fjeirjerij and assert the 404 page appears
- [ ] Assert the Contact support button opens the support widget
- [ ] Assert the Go back home button routes to the homepage

image

### DIFF
--- a/src/pages/404.astro
+++ b/src/pages/404.astro
@@ -1,0 +1,39 @@
+---
+import BaseLayout from '../layouts/BaseLayout.astro';
+---
+
+<BaseLayout>
+  <div class="px-16 py-8 flex flex-col items-center space-y-6 h-full justify-center">
+    <div class="flex flex-col items-center space-y-2">
+      <p class="text-content-accent type-body-2 font-bold">
+        404
+      </p>
+      <h1 class="type-headline-2 text-gray-900">
+        Page not found
+      </h1>
+      <p class="type-body-2 text-content-tertiary text-center">
+        Sorry, we couldn’t find the page you’re looking for.
+      </p>
+    </div>
+    <div class="flex space-x-2 items-center">
+      <a href="/">
+        <div
+          class="btn bg-interactive-primary border rounded-md"
+        >
+          <span class="type-caption-2 text-content-inverse-primary">
+            Go back home
+          </span>
+        </div>
+      </a>
+      <button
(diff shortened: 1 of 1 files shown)


Neeraj-Ghodla (approved):

MeganSteenkamp:
@AndyClifford would be good to get a design review on this ❤️

src/pages/404.astro
```diff
@@ -0,0 +1,39 @@
...
+import BaseLayout from '../layouts/BaseLayout.astro';
+---
+
+<BaseLayout>
+  <div class="px-16 py-8 flex flex-col items-center space-y-6 h-full justify-center">
+    <div class="flex flex-col items-center space-y-2">
+      <p class="text-content-accent type-body-2 font-bold">
+        404
+      </p>
+      <h1 class="type-headline-2 text-gray-900">
```
MeganSteenkamp, line 11: I think this should be a custom Centrapay colour token next time 👀

src/pages/404.astro
```diff
@@ -0,0 +1,39 @@
...
+        >
+          <span class="type-caption-2 text-content-inverse-primary">
+            Go back home
+          </span>
+        </div>
+      </a>
+      <button
+        id="freshdesk"
+        type="button"
+        class="btn bg-interactive-tertiary border rounded-md"
```
MeganSteenkamp, line 31: @AndyClifford How does this work to open the freshdesk? I'm guessing it's from the id but ids should be unique on a page

AndyClifford: Yeah it is from the id - this has broken the original button when the 404 page is visible. Will look into this now

src/pages/404.astro
AndyClifford: Do you have a suggestion for which one to use? Looks like there's a few cases of `text-gray-*` so maybe we need some new classes to capture these

PR #800, opened 2023-10-09
♻️ Refactor side navigation

MeganSteenkamp:
In a follow up PR, we want to be able to have section titles in our side navigation. Our current navigation was done with conditional logic recursing over the same component. This makes it difficult to know what level a heading is unless you have a counter counting the levels of recursion 🤔

Changes:
* Change heading colours to meet accessibility standards (make small text darker).
* Make level 2 headings a consistent font and size, regardless of whether it is a disclosure button.
* Correct semantic HTML for a menu. See reference: https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/#sc_label
* Enable disclosures to open with enter key. See reference: https://headlessui.com/vue/disclosure#keyboard-interaction

Task: https://www.notion.so/centrapay/Sidebar-top-level-non-collapsible-sections-should-be-bigger-884638950dea4dd4940f7ae76addd238?pvs=4

Design review: https://centrapay.slack.com/archives/C04B93QUDSM/p1696825071503949

Test plan: Expect no breaking changes to existing navigation.


src/utils/containsPagePath.js
```diff
@@ -0,0 +1,8 @@
+export default function containsPagePath({ navigationItem, path }) {
```
tom-blake, line 1: @MeganSteenkamp Add tests for this utility

src/components/SecondaryNavDisclosure.vue
```diff
@@ -0,0 +1,59 @@
+<template>
```
tom-blake, line 1: @MeganSteenkamp Suggestion: Refer to these as Sidebar disclosures. "Nav" can be misinterpreted for the navbar along the top of the page.

src/components/SecondaryNavDisclosure.vue
```diff
@@ -0,0 +1,59 @@
...
+  <Disclosure
```
tom-blake, line 2: @MeganSteenkamp Suggestion: There's a fair bit of overlap between PrimaryNavDisclosureButton and this component. Suggest looking at using a single component or a base one

tom-blake (approved):

PR #803, opened 2023-10-11
Endpoint component

AndyClifford:
This change allows us to document api endpoints using markdoc using the same layout as seen in the [Protocol Tailwind UI Example](https://protocol.tailwindui.com/contacts#list-all-contacts).

There are issues with the styling such as the h2 header being underlined but we are just focussed on getting the layout correct then will followup with styling later once we have aligned with the designers. The code block styling needs updating too.

**Test Plan:**
- [ ] Deploy to development
- [ ] Visit http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/api/media-uploads/ and assert the endpoints are rendered correctly

**Desktop:**

image

**Mobile:**

image


src/components/Column.astro
```diff
@@ -0,0 +1,9 @@
+
+---
+const { sticky = false } = Astro.props;
+const className = sticky ? '[&>:first-child]:mt-0 [&>:last-child]:mb-0 xl:sticky xl:top-24' : '[&>:first-child]:mt-0 [&>:last-child]:mb-0';
```
AndyClifford, line 4: Is there a better way to do this ?

src/components/Code.astro
AndyClifford, line 1: **Note:** Removing `client:load` from the ProseCode.astro component caused existing code blocks to look like this:

image

These guides are published to production so for now I have created a new Code.astro component to be used by markdoc until we figure out what is going on with the styling here 👍

MeganSteenkamp: @AndyClifford yup let's make a note of this in the [Style Code Blocks](https://www.notion.so/centrapay/Style-code-blocks-dec29e7eb9474da0b2efaa90546e1dba?pvs=4) task ✅

src/components/Column.astro
MeganSteenkamp: @AndyClifford yes you can just have a ternary operator inline:
```
:class="sticky ? '[&>:first-child]:mt-0 [&>:last-child]:mb-0 xl:sticky xl:top-24' : '[&>:first-child]:mt-0 [&>:last-child]:mb-0'"
```

src/components/Endpoint.astro
```diff
@@ -0,0 +1,28 @@
...
+  emerald: 'ring-emerald-300 dark:ring-emerald-400/30 bg-emerald-400/10 text-emerald-500 dark:text-emerald-400',
+  sky: 'ring-sky-300 bg-sky-400/10 text-sky-500 dark:ring-sky-400/30 dark:bg-sky-400/10 dark:text-sky-400',
+  amber: 'ring-amber-300 bg-amber-400/10 text-amber-500 dark:ring-amber-400/30 dark:bg-amber-400/10 dark:text-amber-400',
+  rose: 'ring-rose-200 bg-rose-50 text-red-500 dark:ring-rose-500/20 dark:bg-rose-400/10 dark:text-rose-400',
+}
+---
+
+<div>
+  <div class="flex items-center gap-x-3">
+    <span class=`font-mono text-[0.625rem] font-semibold leading-6 rounded-lg px-1.5 ring-1 ring-inset + ${colorStyles[color]}`>{ method }</span>
```
MeganSteenkamp, line 21: @AndyClifford I think we should remove all font classes so the font is consistent. We can sort this later with our typography styling.
(I see I accidentally added an instance of `font-mono` last week - I'll remove it 🤠)

src/components/Row.astro
MeganSteenkamp, line 1: Can we remove our existing `Grid` component for these columns and rows?

src/content/api/media-uploads.mdoc
```diff
@@ -44,3 +44,84 @@ When the Media Upload was updated.
 The User or API Key that updated the Media Upload.
 {% /property %}
 {% /properties %}
+
+{% endpoint method="POST" path="/api/media-uploads" %}
+## Create a presigned URL for Media Upload EXPERIMENTAL
```
MeganSteenkamp, line 49: We might want to check this with Junjie - I think we will want to use badges when the route is experimental 🤠

src/content/api/media-uploads.mdoc
```diff
@@ -44,3 +44,84 @@ When the Media Upload was updated.
...
+{% row %}
+{% column %}
+### Required Fields
+{% properties %}
+{% property name="accountId" type="string" %}
+The Media Upload's owning Centrapay Account id.
```
MeganSteenkamp, line 55: Comment: I wonder if we could define the fields, types and descriptions in one place so that we don't need so much repetition. We would have to flag which routes they are used in though 🤔

src/components/Column.astro
alexBinns: Is there a way that we don't have to have such a long line of code?

alexBinns:
What does it look like in mobile view?

src/components/Column.astro
AndyClifford: Cheers - seems like astro has slightly different syntax to vue but got it working with a parameterized string

src/components/Code.astro
MeganSteenkamp: @AndyClifford I just realised the `client:load` means that the copy button works so I think we need this. It also looks like there is no syntax highlighting so I don't think this is correct 🤔

src/content/api/media-uploads.mdoc
```diff
@@ -44,3 +44,84 @@ When the Media Upload was updated.
...
+  -H "X-Api-Key: $api_key" \
+  -H "Content-Type: application/json" \
+  -d '{
+    "accountId": "Jaim1Cu1Q55uooxSens6yk",
+    "mimeType": "image/png",
+    "fileName": "image.png"
+  }'
+```
+
+```json
```
MeganSteenkamp, line 81: @AndyClifford This is the wrong example response 🤠

src/content/api/media-uploads.mdoc
AndyClifford: Looks like its still being used by the farmlands connection page - do you think we should replace it with Rows and Columns in this PR?

src/components/Row.astro
AndyClifford: Looks like its still being used by the farmlands connection page - do you think we should replace it with Rows and Columns in this PR?

src/content/api/media-uploads.mdoc
AndyClifford: Yeah I agree

AndyClifford:
> What does it look like in mobile view?

image

src/content/api/media-uploads.mdoc
```diff
@@ -44,3 +44,69 @@ When the Media Upload was updated.
 The User or API Key that updated the Media Upload.
 {% /property %}
 {% /properties %}
+
+{% endpoint method="POST" path="/api/media-uploads" %}
+## Create a presigned URL for Media Upload EXPERIMENTAL
+{% row %}
+{% column %}
+### Required Fields
```
MeganSteenkamp, line 52: @AndyClifford I think this should now say `Attributes` and the fields should have the required badge on them.
See design: https://www.figma.com/file/IgRlQWJTyGoZhlF4fVKD6B/Centrapay-Docs-(Tailwind)?type=design&node-id=1691-2617&mode=design&t=50p3UEnXzuaCVT5Q-4

src/components/Code.astro
MeganSteenkamp: @AndyClifford it looks like if you have an explicit fence tag we get the syntax highlighting so not sure what's happening there 🤔

MeganSteenkamp:
@AndyClifford we should get some input on the Jesters here too 🤡

src/content/api/media-uploads.mdoc
AndyClifford: I've added a task for styling experimental endpoints. I'd say we want a badge for this which is not part of the anchor tag so we don't break anchor tags when they come out of experimental phase

src/components/Code.astro
AndyClifford: Yeah I think this is a bug - we shouldn't need a custom tag as well as the ``` syntax. 

Ideally using ``` renders a custom code block component with shiki styling

src/components/ProseCode.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div
-    class="my-8 rounded-lg"
+    class="mb-8 rounded-lg"
```
AndyClifford, line 3: This was messing up the layout so removed the margin from the top. I have checked that this doesn't affect existing published documentation.

AndyClifford:
> @AndyClifford we should get some input on the Jesters here too 🤡

@MeganSteenkamp I have checked in with Junjie and he is happy with the current layout

src/components/Code.astro
```diff
@@ -0,0 +1,3 @@
+<div>
```
AndyClifford, line 1: Created this component to avoid having `div` tags in the markdown files. This component will be filled out once we add styling as per the [Figma design](https://www.figma.com/file/IgRlQWJTyGoZhlF4fVKD6B/Centrapay-Docs-(Tailwind)?type=design&node-id=1654-8412&mode=design&t=AEMFKFrkAbgapcAS-0)

MeganSteenkamp: @AndyClifford what styling would end up here?
We could perhaps rename the component to be `CodeColumn` and have the `<Column sticky>` tags here.

AndyClifford: Yeah sounds good. The Protocol example uses `CodeGroup`?

MeganSteenkamp: @AndyClifford yeah perhaps let's rename to `CodeGroup` 👍

AndyClifford: Do you think we will always have code examples i.e. need two columns? If not maybe having the code group responsible for creating the second column is the right play.

MeganSteenkamp: @AndyClifford I think we always would have code examples for an endpoint.

MeganSteenkamp (approved):

PR #808, opened 2023-10-13
Add experimental flag

alexBinns:
This PR adds a badge astro component.

image
image
image

### DIFF
--- a/src/components/Badge.astro
+++ b/src/components/Badge.astro
@@ -1,0 +1,21 @@
+---
+
+const { type = "default", text, link } = Astro.props;
+const badgeText = text ? text : type;
+
+const coloursMap = {
+  experimental: 'bg-surface-accent-light text-content-accent border-outline-opaque',
+  GET: 'border-green-400 bg-green-100 text-content-primary',
+  POST: 'border-cyan-400 bg-cyan-100 text-content-primary',
+  PUT: 'border-orange-400 bg-orange-100 text-content-primary',
+  DELETE: 'border-red-400 bg-rose-100 text-content-primary',
+  default: 'text-content-primary border-outline-opaque bg-surface-tertiary'
+}
+
+const badgeClass =`${coloursMap[type]} type-caption-2 px-2 py-0.5 rounded-full border self-center`
+
+---
+{link ?
+  <a class={badgeClass} href={link}>{badgeText}</a> :
+  <span class={badgeClass}>{badgeText}</span>
+}
--- a/src/components/Property.astro
+++ b/src/components/Property.astro
@@ -1,9 +1,11 @@
 
 ---
+import Badge from '../components/Badge.astro';
+
 const { name, type } = Astro.props;
 const customDataTypes = ['timestamp', 'bignumber', 'monetary', 'crn', 'location', 'phonenumber'];
 const formattedType = type.toLowerCase();
-const isLink = customDataTypes.includes(formattedType);
+const link = customDataTypes.includes(formattedType) ? '/api/data-types#' + formattedType :undefined;
 ---
 
 <li class="m-0 px-0 py-4 first:pt-0 last:pb-0">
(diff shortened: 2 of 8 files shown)


MeganSteenkamp:
@alexBinns we should get a design review on this 🙌 
How does it look when it's on an endpoint title?

alexBinns:
> @alexBinns we should get a design review on this 🙌 How does it look when it's on an endpoint title?

Yea I have posted in design-crit for jesters review

alexBinns:
> @alexBinns we should get a design review on this 🙌 How does it look when it's on an endpoint title?

It looks the same pretty much. I can rebase off Andy's PR and add in there?

AndyClifford:
Looks great 🚀 

- How do we plan on using this in combination with a native markdown h2 i.e. `## Endpoint Title ...`
- Do you think it makes more sense to use a component which encapsulates the h2 using html? If so how will our site nav and/or TOC handle these?

alexBinns:
> @alexBinns we should get a design review on this 🙌 How does it look when it's on an endpoint title?

image
image

src/components/Experimental.astro
```diff
@@ -0,0 +1,3 @@
+<span class="bg-surface-accent-light text-content-accent text-xs px-2 py-0.5 rounded-full">
```
tom-blake, line 1: @alexBinns Suggestion: Create a generic badge component with a prop called "variant" rather than a new component for each variant of badge.

The variant prop can accept semantic badge names (experimental, required, deprecated etc) and encapsulate all styling required for each variant 👍

alexBinns: I have done that now.

src/components/Badge.astro
```diff
@@ -0,0 +1,18 @@
...
+  GET: 'border-green-400 bg-green-100 text-content-primary',
+  POST: 'border-cyan-400 bg-cyan-100 text-content-primary',
+  PUT: 'border-orange-400 bg-orange-100 text-content-primary',
+  DELETE: 'border-red-400 bg-rose-100 text-content-primary',
+  default: 'text-content-primary border-outline-opaque bg-surface-tertiary'
+}
+
+---
+
+<span class=`${ coloursMap[type] } inline-flex text-xs px-2 py-0.5 rounded-full border`>
```
MeganSteenkamp, line 16: @alexBinns are we missing a text style here? The design looks like it has `type-caption-2`

src/components/Property.astro
```diff
@@ -12,7 +14,7 @@ const isLink = customDataTypes.includes(formattedType);
       Name
     </dt>
     <dd>
-      <code>{ name }</code>
+      <Badge text={name}/>
```
MeganSteenkamp, line 17: Should this be on the type? https://www.figma.com/file/IgRlQWJTyGoZhlF4fVKD6B/Centrapay-Docs-(Tailwind)?type=design&node-id=1718-454&mode=design&t=ZPbJbXOw7WVXAitK-4

src/components/Badge.astro
```diff
@@ -0,0 +1,18 @@
+---
+
+const { type = "default", text } = Astro.props;
+
+const coloursMap = {
```
MeganSteenkamp, line 5: Let's add a required badge in another increment 🙌

src/components/Property.astro
alexBinns: Yea I have changed this

alexBinns:
I have updated it to a flex that will drop down to a new line instead of hover at the right side screenshots below. I spoke to Junjie. He said it is preferable to having it hover on the right side.
image
image
image

src/components/Badge.astro
```diff
@@ -0,0 +1,21 @@
...
+  experimental: 'bg-surface-accent-light text-content-accent border-outline-opaque',
+  GET: 'border-green-400 bg-green-100 text-content-primary',
+  POST: 'border-cyan-400 bg-cyan-100 text-content-primary',
+  PUT: 'border-orange-400 bg-orange-100 text-content-primary',
+  DELETE: 'border-red-400 bg-rose-100 text-content-primary',
+  default: 'text-content-primary border-outline-opaque bg-surface-tertiary'
+}
+
+const badgeClass =`${coloursMap[type]} type-caption-2 text-xs px-2 py-0.5 rounded-full border self-center`
```
AndyClifford, line 15: Suggestion: We don't need the `text-xs` class as it already exists on type-caption-2

AndyClifford (approved):

PR #810, opened 2023-10-13
✨ Add H2 headings to side navigation

MeganSteenkamp:
Add H2 headings into left hand side navigation menu.

Related thread design review: https://centrapay.slack.com/archives/C04B93QUDSM/p1697156730162259
Story: https://www.notion.so/centrapay/Migrate-API-Documentation-c6d64de29a2245358a5e9b7eed5edbcc?pvs=4

Test plan: Expect to see H2 headings for the current page in the left hand side navigation. Expect current headings to have highlighting.

image


tsconfig.json
MeganSteenkamp, line 1: This was giving me type errors in Astro files where I think we can just plain js. Has anyone else been having this issue?

src/__tests__/__snapshots__/links.integration.test.js.snap
MeganSteenkamp, line 1: This has added in references to all of the anchor tags 👀

tsconfig.json
AndyClifford: Yeah something like this was happening to me when we were playing around with that shiki transform stuff inside an astro component.

AndyClifford (approved):

alexBinns (approved):

PR #813, opened 2023-10-17
Migrate accounts page

AndyClifford:
Migrated the accounts API reference page to the new astro site in draft mode. Will deploy each page to dev for people to have a look at instead of posting screenshots as the pages are quite long

**Ticket:** https://www.notion.so/centrapay/Migrate-API-Documentation-c6d64de29a2245358a5e9b7eed5edbcc?pvs=4

**Test Plan:**
- [x] Deploy to dev
- [x] Go to http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/api/accounts/ and assert the page is rendered correctly
- [ ] Go to https://docs.centrapay.com/api/accounts/ and assert the legacy page is still rendered

### DIFF
--- a/src/content/api/accounts.mdx
+++ b/src/content/api/accounts.mdx
@@ -1,0 +1,263 @@
+---
+title: Accounts
+description: Account model and related endpoints
+draft: true
+nav:
+  path: API
+  order: 7
+---
+import Properties from '../../components/Properties.astro';
+import Property from '../../components/Property.astro';
+import Endpoint from '../../components/Endpoint.astro';
+import CodeGroup from '../../components/CodeGroup.astro';
+
+An Account represents a permission boundary around Centrapay resources.
+Accounts can have [API Keys](https://docs.centrapay.com/api/api-keys) and [Account Memberships](https://docs.centrapay.com/api/account-memberships) which grant access to the resources.
+
+Accounts are classified as either "individual" or "org". Individual accounts
+can only have a single member and Centrapay users can only be a member of a single
+individual account.
+
+## Account Model
+
+### Attributes
+
+<Properties>
+  <Property name="id" type="string">
+    The unique identifier.
+  </Property>
+
+  <Property name="type" type="string">
+    Account type, must be either 'org' or 'individual'.
+  </Property>
+
+  <Property name="name" type="string">
+    The display name of the Account.
+  </Property>
+
+  <Property name="region" type="string">
+    The region that the Account will operate in.
+  </Property>
+
+  <Property name="test" type="boolean">
+    A flag which is only present if the Account is for testing.
+  </Property>
+
+  <Property name="createdAt" type="timestamp">
+    When the Account was created.
+  </Property>
+
+  <Property name="modifiedAt" type="timestamp">
+    When the Account was updated.
+  </Property>
+
+  <Property name="createdBy" type="crn">
+    The User or API Key that created the Account.
+  </Property>
+
+  <Property name="modifiedBy" type="crn">
+    The User or API Key that updated the Account.
+  </Property>
+
+  <Property name="subscriptions" type="array">
+    A list of [Subscriptions](#subscription-model) on the Account.
+  </Property>
+</Properties>
+
+## Subscription Model
+
+### Attributes
+
+<Properties>
+  <Property name="name" type="string">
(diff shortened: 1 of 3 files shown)


AndyClifford:
@MeganSteenkamp @alexBinns This is what the get account endpoint looks like with no attributes section. What do you guys think? I think it looks fine and means devs don't have to worry about adding `No Attributes` for these endpoints. Happy to keep as is though

image

src/content/api/accounts.mdx
```diff
@@ -0,0 +1,263 @@
...
+import CodeGroup from '../../components/CodeGroup.astro';
+
+An Account represents a permission boundary around Centrapay resources.
+Accounts can have [API Keys](/api/api-keys/) and [Account Memberships](/api/account-memberships/) which grant access to the resources.
+
+Accounts are classified as either "individual" or "org". Individual accounts
+can only have a single member and Centrapay users can only be a member of a single
+individual account.
+
+## Account Model
```
AndyClifford, line 21: The media uploads page has `Model` but that doesn't work very well for this page as we have two models. Any suggestions here. We could just have `Account` and `Subscription`

src/content/api/accounts.mdx
```diff
@@ -0,0 +1,263 @@
...
+  path: API
+  order: 7
+---
+import Properties from '../../components/Properties.astro';
+import Property from '../../components/Property.astro';
+import Endpoint from '../../components/Endpoint.astro';
+import CodeGroup from '../../components/CodeGroup.astro';
+
+An Account represents a permission boundary around Centrapay resources.
+Accounts can have [API Keys](/api/api-keys/) and [Account Memberships](/api/account-memberships/) which grant access to the resources.
```
AndyClifford, line 15: @MeganSteenkamp linkinator is failing because `/api/account-memberships/` is not a valid link yet. What do you think we should do here?

I think we should only validate links on the production site - this allows us to migrate each page with a single PR. Links will then be validated in the final PR which publishes every page.

The other option is to leave these links and come back to them for each page but I think this will take longer.

MeganSteenkamp: @AndyClifford I think we can make them absolute links to pass the pipeline and then migrate them all at the end.

alexBinns:
> tter as long as there is an endpoint description. It also means devs don't have to worry about adding `No Attributes` for these endpoints. Happy to keep as is though.

I feel for consitancy that we should specify no attributes and it will be less likely to cause confusion

src/content/api/accounts.mdx
alexBinns: I agree with Megan. This way we won't miss any links resulting in what we are trying to avoid with the check.

AndyClifford: I can see the ProseA custom component has logic to determine whether or not the link is for the api reference

`const isApiReferenceLink = props.href.startsWith('https://docs.centrapay.com/api');`

Let's leave the validation as is and just put in the full path. We can open a PR at the end to change this logic allowing our links to be relative.

src/content/api/accounts.mdx
alexBinns: Maybe we should name the models rather than have just model. Do we even need a model as it is just an array. Could we just add to the account model?

AndyClifford: I'm not sure what you mean here. At the moment we have two named models - `Account Model` and `Subscription Model`. But these could be changed to `Account` and `Subscription`. 

Subscriptions is just an array but we have other use cases where the sub-model is much larger and so having a single model could get messy

alexBinns: Sorry I read over my comment and it was a bit confusing. I was esentiallly agreeing with you to just have the name of the model and leave off the model.
The suggestion for the subscription model was just for this particular scenario.

AndyClifford: Sweet as - I think `Account Model` reads better in the navigation bar 🤔

MeganSteenkamp (approved):

src/content/api/accounts.mdx
MeganSteenkamp: @AndyClifford suggestion: We have `Model` as a H2 and then H3 `Account` and `Subscription`

PR #815, opened 2023-10-17
Add required badge

alexBinns:
This PR adds a required badge. 

Testing:
Assert that the attributes on media uploads page have required badge

### DIFF
--- a/src/components/Property.astro
+++ b/src/components/Property.astro
@@ -20,7 +20,9 @@ const link = customDataTypes.includes(formattedType) ? '/api/data-types#' + form
     <dd>
       <Badge text={formattedType} link={link}>
     </dd>
-    { required ? <dd> <Badge type="required"> </dd> : undefined }
+    { required &&
+      <dt class="sr-only">Required</dt>
+      <dd> <Badge type="required"> </dd> }
     <dt class="sr-only">
       Description
     </dt>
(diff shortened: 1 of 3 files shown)


src/components/Property.astro
```diff
@@ -17,9 +17,10 @@ const link = customDataTypes.includes(formattedType) ? '/api/data-types#' + form
       {name}
     </dd>
     <dt class="sr-only">Type</dt>
-    <dd class="font-mono text-xs text-zinc-400 dark:text-zinc-500">
+    <dd>
```
MeganSteenkamp, line 20: @alexBinns each <dd> tag should have a <dt> with it.
If we don't want it visible we can give it the class `sr-only`.
You can also make the conditional logic simpler.

Example:
```
{ required && 
  <dt class="sr-only">Required</dt>
  <dd><Badge type="required"></dd>
}
```

src/content/api/media-uploads.mdx
```diff
@@ -66,15 +66,15 @@ import Badge from '../../components/Badge.astro';
 
   ### Attributes
   <Properties>
-    <Property name="accountId" type="string">
+    <Property name="accountId" type="string" required="true">
```
MeganSteenkamp, line 69: Suggestion: Can you just pass `required` without explicitly setting `true`?

AndyClifford (approved):

PR #823, opened 2023-10-18
Migrate asset types page

alexBinns:
This PR migrates the asset types page.
I have also updated the table to remove the test assets and instead replaced with a flag on for assets that support test. LMK what you think.

### DIFF
--- a/src/content/api/asset-types.mdx
+++ b/src/content/api/asset-types.mdx
@@ -1,0 +1,36 @@
+---
+title: Asset Types
+description: Asset types supported for payment
+draft: true
+nav:
+  path: API
+  order: 10
+---
+
+The following table describes the Asset Types supported for payments.
+
+The Category column refers to the Centrapay asset type representation if
+applicable. When blank, the Asset is not managed by a Centrapay Account.
+
+|   Asset Type    |                Description                |                          Category                                  |           Currencies          |    Flags    |
+| :---------------|:----------------------------------------- | :----------------------------------------------------------------- | :---------------------------- | :---------- |
+| bitcoin         | [Bitcoin](https://bitcoin.org/en/)        |                                                                    | `NZD` `AUD`                   | ⚡️          |
+| cca.coke        | Coke tokens                               | [Token](https://docs.centrapay.com/api/assets#tokens-experimental) | `NZD`                         | ⚡️ 🧪 🚫    |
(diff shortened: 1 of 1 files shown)


src/content/api/asset-types.mdx
```diff
@@ -0,0 +1,34 @@
+---
+title: Supported Asset Types
```
MeganSteenkamp, line 2:
```suggestion
title: Asset Types 
```

IMO this looks better in the nav bar ✨

MeganSteenkamp:
@alexBinns I think I prefer the existing table. IMO it looks like you can use `bitcoin.main` as a test asset the way it is laid out now. 
We are also missing `uplinkapi.test` in this table.

alexBinns:
> @alexBinns I think I prefer the existing table. IMO it looks like you can use `bitcoin.main` as a test asset the way it is laid out now. We are also missing `uplinkapi.test` in this table.

We are not gaining anything from having the test and the main separate and I feel it makes the table harder to read

alexBinns:
> @alexBinns I think I prefer the existing table. IMO it looks like you can use `bitcoin.main` as a test asset the way it is laid out now. We are also missing `uplinkapi.test` in this table.

That is what the T is for to indicate what assets can be used with test

src/content/api/asset-types.mdx
```diff
@@ -0,0 +1,36 @@
+---
+title: Asset Types
+description: Asset types supported for payment
+draft: true
+nav:
+  path: API
+  order: 10
```
MeganSteenkamp, line 7: Should this be 12? How are you counting this?
It's the 12 based on the existing docs site.

alexBinns: Updated

src/content/api/asset-types.mdx
```diff
@@ -0,0 +1,38 @@
...
+---
+
+The following table describes the Asset Types supported for payments.
+
+Asset types support live and test assets unless specified by supplying the correct suffix e.g. `centrapay.nzd.main` for live and `centrapay.nzd.test` for test.
+
+The Category column refers to the Centrapay asset type representation if
```
MeganSteenkamp, line 14:
```suggestion
The Category column refers to the Centrapay [Asset category](/api/assets/) representation if
```

MeganSteenkamp (approved):

src/content/api/asset-types.mdx
alexBinns: I have removed this line. It was a previous iteration of the table. I have now updated to have it mentioned in the legend instead.

PR #824, opened 2023-10-18
🐛 Fix endpoint layout

MeganSteenkamp:
Remark plugin was causing sections to be rendered before slots were used, which was affecting our layout. Change to use a rehype plugin so HTML is transformed instead of transforming MDX.

Test plan: Expect guides layout and table of contents to be unaffected.


package.json
```diff
@@ -16,6 +16,7 @@
     "@babel/eslint-parser": "^7.14.7",
     "@babel/eslint-plugin": "^7.14.5",
     "@fontsource/inter": "^4.5.15",
+    "@hbsnow/rehype-sectionize": "^1.0.7",
```
MeganSteenkamp, line 19: There is a more popular package `[rehype-section](https://www.npmjs.com/package/@agentofuser/rehype-section)`  but downloads aren't working for it.

yarn.lock
```diff
@@ -820,6 +820,15 @@
   resolved "https://registry.yarnpkg.com/@fontsource/inter/-/inter-4.5.15.tgz#eed1873d68755d3b52d6fcfcfa3493118430a512"
```
tom-blake, line 820: @MeganSteenkamp Did you use `yarn remove` here? I can't see it being removed from the lock file

MeganSteenkamp: Thanks @tom-blake have updated ❤️

AndyClifford (approved):

PR #834, opened 2023-10-19
⏩ Migrate Wallets API documentation

MeganSteenkamp:
Story: https://www.notion.so/centrapay/Migrate-API-Documentation-c6d64de29a2245358a5e9b7eed5edbcc?pvs=4

Test plan: Expect to see migrated page in preview mode.

### DIFF
--- a/src/content/api/wallets.mdx
+++ b/src/content/api/wallets.mdx
@@ -1,0 +1,285 @@
+---
+title: Wallets
+description: Wallet models and related endpoints
+draft: true
+nav:
+  path: API
+  order: 15
+---
+import Properties from '../../components/Properties.astro';
+import Property from '../../components/Property.astro';
+import Badge from '../../components/Badge.astro';
+import Error from '../../components/Error.astro';
+import Endpoint from '../../components/Endpoint.astro';
+import CodePanel from '../../components/CodePanel.astro';
+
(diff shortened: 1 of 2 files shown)


src/content/api/wallets.mdx
```diff
@@ -0,0 +1,285 @@
...
+  No attributes.
+
+  <CodePanel slot="code-examples" title="Request" method="GET" path="/api/wallets/{walletId}/transactions">
+    ```bash
+    curl https://service.centrapay.com/api/wallets/WRhAxxWpTKb5U7pXyxQjjY/transactions \
+      -H "X-Api-Key: $api_key"
+    ```
+  </CodePanel>
+
+  <CodePanel slot="code-examples" title="Response">
```
MeganSteenkamp, line 267: There are a lot of examples in the existing documentation. I was going to do multiple items but JSON doesn't support comments so could be confusing.

AndyClifford: Can we include all of the examples as separate code panels with the example name as the title?

AndyClifford (approved):

src/content/api/wallets.mdx
MeganSteenkamp: @AndyClifford we could but I had discussed this with Tom and the jesters and the patterns was going to be one response code panel per endpoint, with the idea that our guides should be better.

PR #850, opened 2023-10-24
Add legacy redirects

AndyClifford:
We are about to remove the legacy Jekyll site which contains the redirect configuration for a lot of pages. These have been moved to the astro site to avoid breaking links.

**Ticket Link:** https://www.notion.so/centrapay/Release-new-API-layout-in-production-9e3974fee0d140eda47b6996e04f829c?pvs=4

**Test Plan:** 
- [ ] Expect the newly added redirects to work on the astro site.


MeganSteenkamp:
@AndyClifford what will the behaviour in production be while we still have the draft pages?

AndyClifford:
> @AndyClifford what will the behaviour in production be while we still have the draft pages?

Unsure - I think the trailing slash on the `redirect to` will mean we won't be routed to the legacy site and instead see a 404. I'd be keen to see what actually happens here as we don't have a way of verifying this locally. But I think I'll just merge this PR with the publish one to keep things simple.

AndyClifford:
have included this commit in the publish PR to avoid momentarily breaking redirects

PR #851, opened 2023-10-24
Publish Draft Pages

AndyClifford:
We have finished migrating all of the legacy API documentation to the Astro site as draft pages. This change will publish the content in production. Consumers of our docs will not be able to access the legacy content 🥳 .

**Notes:** 
- There are a couple of outstanding content migration PR's. Will update this PR once they have been merged as draft pages.
- Have migrated the out standing legacy redirects to avoid breaking redirects

**Ticket Link:** https://www.notion.so/centrapay/72a96dfb4d6e451c994e786b04b9747e?pvs=4

**Test Plan:**
- [ ] Post in #development asking to freeze new changes so the jekyll site can be removed
- [ ] Assert there is only a single `API` section in the side navigation.
- [ ] Access each page in the API section to assert there are no unexpected issues.
- [ ] Make sure redirects work as expected
- [ ] Make sure the legacy api reference can not be accessed. Try https://docs.centrapay.com/api/introduction.html


src/content/api/media-uploads.mdx
```diff
@@ -1,10 +1,9 @@
 ---
 title: Media Uploads
 description: Media Uploads API Reference
-draft: true
 nav:
   path: API
-  order: 7
+  order: 25
```
IsaacMurtagh, line 6: This system seems very cumbersome to insert a document in between another document.
Here are some alternatives that may make it easier.

## Use directories, then prepend files and directories with their order
This is going to reduce the number of documents you need to edit when inserting a document, plus it's far more obvious what documents need to be edited.

## A separate config file which dictates the order
You can have a file that includes an array of file paths in the order you want them displayed. Very easy to change the order, the downside is you need to know how to edit this when adding a new document.

AndyClifford: Yeah I agree its not great 😅 - I'll add a task to address this

AndyClifford: It could be nice if each file declared which file it is `after`

MeganSteenkamp: I agree @IsaacMurtagh I think prepending the files is a nice solution here 💯

AndyClifford: I'm not sure I understand the directories suggestion - if files/directories are prepended with their order won't this require updating all subsequent file/directory names?

alexBinns (approved):

src/content/api/media-uploads.mdx
alexBinns: @AndyClifford I think what he means is give the directory an index and the page within an index as well. For example `src/content/api/1-assets/1-assets.mdx` and `src/content/api/1-assets/2-asset-transfers.mdx` etc

MeganSteenkamp:
@AndyClifford it would be good to have search working before we release this 👀

PR #853, opened 2023-10-25
Don't add attribute and error headings to search index

alexBinns:
This PR makes it so the attributes and errors headings aren't added to the search index.

Testing:
* Deploy to dev
* Search for attributes assert that there are no attributes headings displayed
* Search for errors assert that there are no errors headings displayed

### DIFF
--- a/src/plugins/flexsearch.js
+++ b/src/plugins/flexsearch.js
@@ -56,6 +56,9 @@ async function createFlexsearchIndexData() {
     const vfile = await processor.process(content);
     // Add all other headings and their descriptions to the index
     vfile.data.sections.forEach(({ title, description }) => {
+      if([ 'Attributes', 'Errors' ].includes(title)) {
+        return;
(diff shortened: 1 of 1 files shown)


src/plugins/flexsearch.js
```diff
@@ -56,6 +56,9 @@ async function createFlexsearchIndexData() {
     const vfile = await processor.process(content);
     // Add all other headings and their descriptions to the index
     vfile.data.sections.forEach(({ title, description }) => {
+      if([ 'Attributes', 'Errors' ].includes(title)) {
```
AndyClifford, line 59: Could we just omit all H3 headings?

MeganSteenkamp: I'm not sure about omitting all - an example might be in payment requests we have an H3 which is 'Payment Activity Types'. This would be good to have in our search

PR #858, opened 2023-10-26
🔧 Migrate absolute links & fix anchors

MeganSteenkamp:
- Migrate absolute links
- Fix all invalid anchor tags

Test plan: I have run automated testing on the docs preview site to identify any broken links/ anchors.


public/asset-lifecycle.png
MeganSteenkamp, line 1: @alexBinns this was referenced to the legacy site which will be removed so added it to the new site.

src/components/ProseA.vue
```diff
@@ -28,7 +28,7 @@ const props = defineProps({
   },
 });
 
-const isApiReferenceLink = props.href.startsWith('https://docs.centrapay.com/api');
+const isApiReferenceLink = props.href.startsWith('/api');
```
MeganSteenkamp, line 31: Do we still want this API reference badge?

public/asset-lifecycle.png
MeganSteenkamp, line 1: @alexBinns this was referenced to the legacy site which will be removed so added it to the new site.

src/components/ProseA.vue
alexBinns: I think it might be useful for the guides as the guides were written for people that aren't engineers and they might want to know if they are going to the api. In the api docs we probably don't need it. I'm not sure if there is a way to have it in the guides but not the api docs?

alexBinns (approved):

src/components/ProseA.vue
MeganSteenkamp: Yeah for sure - I'll leave this for now 🙌

PR #864, opened 2023-10-27
Nav order

AndyClifford:
**Ticket Link:** https://www.notion.so/centrapay/Lanterns-d784e8a5f9fa44328a5d0ae1fd85d37e?p=c6d64de29a2245358a5e9b7eed5edbcc&pm=s

**Test Plan:**
- [ ] Assert the nav is rendered correctly and pages/anchors are accessible by clicking the nav items.

**Notes:**
- Some legacy nav groups (e.g. Settlements) only had one page in them so these have been removed.
- I don't like how the nav order is determined by both the pages and the sideNavConfig. This works for now but I think it needs revisiting in the follow up epic.

**Legacy Nav:**

image

**New Nav:**

image

### DIFF
--- a/src/content/api/accounts.mdx
+++ b/src/content/api/accounts.mdx
@@ -3,8 +3,8 @@ title: Accounts
 description: Account model and related endpoints
 draft: true
 nav:
-  path: API
-  order: 7
+  path: API/Accounts
+  order: 1
 ---
 import Properties from '../../components/Properties.astro';
 import Property from '../../components/Property.astro';
--- a/src/navigation/Navigation.js
+++ b/src/navigation/Navigation.js
@@ -2,6 +2,7 @@ import Page from './Page';
 
 class NavGroup {
   constructor(props) {
+    this.order = props.order,
     this.title = props.title;
     this.subTitle = props.subTitle;
     this.children = props.children;
@@ -25,10 +26,11 @@ class NavGroup {
       subTitle: nav.subTitle,
       icon: nav.icon,
       href: nav.href,
+      order: nav.order || 0,
       children: [
         ...childPages,
         ...childNavGroups,
-      ],
+      ].sort((a, b) => a.order - b.order),
     });
   }
 }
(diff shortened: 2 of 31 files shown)


src/navigation/Navigation.js
```diff
@@ -25,10 +26,11 @@ class NavGroup {
       subTitle: nav.subTitle,
       icon: nav.icon,
       href: nav.href,
+      order: nav.order || 0,
       children: [
         ...childPages,
         ...childNavGroups,
-      ],
+      ].sort((a, b) => a.order - b.order),
```
AndyClifford, line 33: This allows pages at the same depth as a nav group to be ordered

src/navigation/sideNavConfig.js
```diff
@@ -22,6 +22,15 @@ const navigationItems = [
     title: 'API',
     subTitle: 'For developers',
     icon: 'Settings',
+    children: [
```
AndyClifford, line 25: It could be better if child nav groups are determined by `nav.path` in the frontmatter instead of in here.

MeganSteenkamp:
@AndyClifford how do you actually click on to the 'Accounts' page?
The click opens the disclosure so I think you need the page again under this OR the click needs to take you to the accounts page.

src/content/api/accounts.mdx
```diff
@@ -2,8 +2,8 @@
 title: Accounts
 description: Account model and related endpoints
 nav:
-  path: API
-  order: 7
+  path: API/Accounts/Sub Accounts
```
MeganSteenkamp, line 5: @AndyClifford how do you actually click on to the 'Accounts' page?
The click opens the disclosure so I think you need the page again under this OR the click needs to take you to the accounts page. Not sure if this is a typo

src/navigation/sideNavConfig.js
MeganSteenkamp: @AndyClifford I agree but I think this is still an improvement 🚀

MeganSteenkamp (approved):

src/content/api/accounts.mdx
AndyClifford: This is a typo left in from some testing

PR #866, opened 2023-10-27
Underline links in badges

AndyClifford:
**Ticket Link:** https://www.notion.so/centrapay/Lanterns-d784e8a5f9fa44328a5d0ae1fd85d37e?p=03bdab510ae84937894da06bb1b5b457&pm=s

**Test Plan:**
- [ ] Assert badges with links in them are underlined and still link to the relevant data type.

image

### DIFF
--- a/src/components/Badge.astro
+++ b/src/components/Badge.astro
@@ -25,10 +25,12 @@ const inverseColoursMap = {
 }
 
 const map = inverse ? inverseColoursMap : coloursMap;
-const badgeClass =`${map[type]} type-caption-2 px-2 py-0.5 rounded-full border self-center`
+const badgeClass =`${map[type]} type-caption-2 px-2 py-0.5 rounded-full border`;
 
 ---
-{link ?
-  <a class={badgeClass} href={link}>{badgeText}</a> :
-  <span class={badgeClass}>{badgeText}</span>
-}
+<span class="not-prose flex items-center">
+  {link ?
+    <a class={badgeClass + ' underline'} href={link}>{badgeText}</a> :
+    <span class={badgeClass}>{badgeText}</span>
+  }
+</span>


src/components/Badge.astro
```diff
@@ -25,10 +25,12 @@ const inverseColoursMap = {
...
 const map = inverse ? inverseColoursMap : coloursMap;
-const badgeClass =`${map[type]} type-caption-2 px-2 py-0.5 rounded-full border self-center`
+const badgeClass =`${map[type]} type-caption-2 px-2 py-0.5 rounded-full border self-center`;
 
 ---
-{link ?
-  <a class={badgeClass} href={link}>{badgeText}</a> :
-  <span class={badgeClass}>{badgeText}</span>
-}
+<div class="not-prose">
```
MeganSteenkamp, line 31: @AndyClifford this is throwing off your alignment if you look at a heading (e.g. http://localhost:4321/api/bank-account-connection-intents#create-bank-account-connection-intent)

Suggested change:
```
const badgeClass =`${map[type]} type-caption-2 px-2 py-0.5 rounded-full border`;

---
<span class="not-prose flex items-center">
  {link ?
    <a class={badgeClass + ' underline'} href={link}>{badgeText}</a> :
    <span class={badgeClass}>{badgeText}</span>
  }
</span>
```

AndyClifford: Do you mean the experimental badge drops down onto the next line?

MeganSteenkamp: image

MeganSteenkamp: @AndyClifford it's not aligned on the vertical axis. I've changed classes in the suggestion too :)

alexBinns (commented):
What is the reason for wrapping it in a span?

AndyClifford:
> What is the reason for wrapping it in a span?

We have to wrap it in something to apply the `not-prose` class. Not 💯 on why span is used instead of a div - maybe so the badges are always inline with whatever they're next to?

alexBinns:
> > What is the reason for wrapping it in a span?
> 
> We have to wrap it in something to apply the `not-prose` class. Not 💯 on why span is used instead of a div

The span was so it would be inline. But maybe because it is a component it could have been a div? Can we not just add not-prose to the `badgeClass`?

MeganSteenkamp:
> > > What is the reason for wrapping it in a span?
> > 
> > 
> > We have to wrap it in something to apply the `not-prose` class. Not 💯 on why span is used instead of a div
> 
> The span was so it would be inline. But maybe because it is a component it could have been a div? Can we not just add not-prose to the `badgeClass`?

"A div element is used for block-level organization and styling of page elements, whereas a span element is used for inline organization and styling." 
Technically you can use either but IMO a span is more syntactically correct.
You need to add `not-prose` one level above.

MeganSteenkamp (approved):

PR #867, opened 2023-10-30
Specify attributes and errors in search index

alexBinns:
This PR appends the context to the attribute and error titles in the search index.

Testing:

- [ ] Search for attributes in the docs site and see that they all have the model or route appended
- [ ] Search for errors in the docs site and see that they all have the route appended

### DIFF
--- a/src/plugins/flexsearch.js
+++ b/src/plugins/flexsearch.js
@@ -8,18 +8,39 @@ import remarkParse from 'remark-parse';
 import { visit } from 'unist-util-visit';
 import { toString } from 'mdast-util-to-string';
 import { findAfter } from 'unist-util-find-after';
+import { findBefore } from 'unist-util-find-before';
 
 function slugify(str) {
   return str.toLowerCase().replace(/\s/g, '-');
 }
 
+function stripBadge(str) {
+  return str.replace(/<Badge.*>/, '').trim();
+}
+
+function formatTitle(node, root) {
+  const nodeString = toString(node);
+
+  if (nodeString == 'Attributes') {
+    const parentHeading = stripBadge(toString(findBefore(root, node, 'heading')));
+    return stripBadge(`${parentHeading} ${nodeString}`);
+  }
+
+  if (nodeString == 'Errors') {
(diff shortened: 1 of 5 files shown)


src/plugins/flexsearch.js
```diff
@@ -8,18 +8,39 @@ import remarkParse from 'remark-parse';
...
+    const parentHeading = stripBadge(toString(findBefore(root, node, 'heading')));
+    return stripBadge(`${parentHeading} ${nodeString}`);
+  }
+
+  if (nodeString == 'Errors') {
+    const headingNode = toString(findBefore(root, findBefore(root, node, 'heading'), 'heading'));
+    return stripBadge(`${headingNode} ${nodeString}`);
+  }
+
+  return stripBadge(nodeString);
```
MeganSteenkamp, line 34: @alexBinns I can still see the badge, here is an example:
image

alexBinns: There was an issue with my regex I have fixed now
image

src/plugins/flexsearch.js
```diff
@@ -40,7 +61,7 @@ async function createFlexsearchIndexData() {
     .use(remarkExtractSections)
     .use(remarkHtml);
 
-  for (const filepath of await promisify(glob)('src/content/**/*.{md,mdx}')) {
+  for (const filepath of await promisify(glob)('src/content/**/api/*.{md,mdx}')) {
```
MeganSteenkamp, line 64: @alexBinns will this remove the ability to search our guides?

alexBinns: Good catch this was for testing. I'll revert this change.

AndyClifford:
@alexBinns I'm not sure what the expected results are supposed to be but here's a screenshot of searching for Attributes locally. 

- The model / route is prepended to the result header ✅ 
- When the result is an endpoint, is there a way we can hide the `<Endpoint ... ` html and just display the method and path? 🤔 
- Are we able to suffix the href of duplicate headings on the same page with `-1`, `-2` etc so they route to the correct place? 🤔 


image

AndyClifford (approved):

PR #868, opened 2023-10-31
Make attributes and errors headings

alexBinns:
This PR converts the attribute name and error name to be headings and adds an id so we can link to them.

### DIFF
--- a/src/components/Badge.astro
+++ b/src/components/Badge.astro
@@ -25,10 +25,10 @@ const inverseColoursMap = {
 }
 
 const map = inverse ? inverseColoursMap : coloursMap;
-const badgeClass =`${map[type]} type-caption-2 px-2 py-0.5 rounded-full border`;
+const badgeClass =`${map[type]} align-text-top type-caption-2 px-2 rounded-full border`;
 
 ---
-<span class="not-prose flex items-center">
+<span class="not-prose flex items-center align-text-top">
   {link ?
     <a class={badgeClass + ' underline'} href={link}>{badgeText}</a> :
     <span class={badgeClass}>{badgeText}</span>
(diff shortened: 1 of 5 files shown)


src/components/Badge.astro
```diff
@@ -25,10 +25,10 @@ const inverseColoursMap = {
 }
 
 const map = inverse ? inverseColoursMap : coloursMap;
-const badgeClass =`${map[type]} type-caption-2 px-2 py-0.5 rounded-full border`;
+const badgeClass =`${map[type]} align-text-top type-caption-2 px-2 rounded-full border`;
```
MeganSteenkamp, line 28: @alexBinns what was the reasoning for changing this?
* I can't see a difference from having the `align-text-top` class.
* We now have less padding on the y axis (`py-0.5`)

src/components/Property.astro
```diff
@@ -8,32 +8,15 @@ const formattedType = type.toLowerCase();
...
-      <dt class="sr-only">Deprecated</dt>
-      <dd> <Badge type="deprecated"> </dd> }
-    <dt class="sr-only">
-      Description
-    </dt>
-    <dd class="w-full flex-none [&>:first-child]:mt-0 [&>:last-child]:mb-0">
+<li class="m-0 px-0 py-4 first:pt-0 last:pb-0 space-y-2">
+  <div class="m-0 flex flex-wrap items-center gap-x-3">
+    <h4 class="my-0" id={name}>{name}</h4>
+      <Badge text={formattedType} link={link} />
```
MeganSteenkamp, line 14: Suggestion: It looks like the nesting here is incorrect i.e. these classes are tabbed in one too many.
We need to add better linting for Astro files to fix this.

MeganSteenkamp (approved):

PR #887, opened 2023-11-21
✨ Generate code snippets

MeganSteenkamp:
Change to use [httpsnippet](https://github.com/Kong/httpsnippet) to generate code snippets.

See preview: https://docs-preview.cp42.click/api/invitations/

Outstanding tasks:
- [ ] Create an ADR for `httpsnippet`


scripts/create-code-snippets.js
MeganSteenkamp, line 1: I'll add testing to this if we like this pattern ✨

MeganSteenkamp:
Outstanding differences:
- Snippets are generated with single-quotes which means you can't use variables for auth like we normally would: https://stackoverflow.com/questions/40852784/how-can-i-use-a-variable-in-curl-call-within-bash-script
- The top banner now has the variables in for the URL i.e. `/api/invitations/code/WIj211vFs9cNACwBb04vQw` instead of `/api/invitations/code/{code}`. I think this is okay but we could manually pass it in how we were before.

package.json
```diff
@@ -4,7 +4,7 @@
     "lint": "eslint --ext .js,.mjs,.ts,.vue .",
     "linkinator": "linkinator dist/ --recurse --verbosity error --skip 'twitter.com|www.farmlands.co.nz|www.instagram.com|yourapp.example.com|centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/api'",
     "dev": "astro dev",
-    "build": "astro build",
+    "build": "./scripts/create-code-snippets.js && astro build",
```
tom-blake, line 7: @MeganSteenkamp Suggestion: Integrate this directly with Astro's build using the `astro:build:start` hook

https://docs.astro.build/en/reference/integrations-reference/#astrobuildstart

scripts/create-code-snippets.js
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const yaml = require('js-yaml');
+const fs = require('fs');
+const path = require('path');
+
+function transformYamlToHAR(data) {
```
tom-blake, line 9: @MeganSteenkamp Why did you choose YAML?

scripts/create-code-snippets.js
tom-blake: @MeganSteenkamp Change: We should add tests to validate code snippet files for developers

src/components/RequestCodePanel.astro
```diff
@@ -0,0 +1,23 @@
...
+console.log({ request });
+const path = request.uriObj.pathname;
+const method = request.method;
+
+const options = { short: true, prettifyJson: true };
+const output = snippet.convert('shell', 'curl', options) as string;
+
+---
+
+{output &&
```
tom-blake, line 19: @MeganSteenkamp Suggestion: Fail the build if output is undefined

src/content/api/_code-snippets/src/invitations-create.yaml
```diff
@@ -0,0 +1,12 @@
+method: POST
```
tom-blake, line 1: @MeganSteenkamp Why did you choose YAML?

tom-blake: Did you consider JS modules? They'd be easier to manipulate and validate

```js
export default {
  path: '/api/invitations',
  headers: {
    X-Api-Key: '$api-key',
    Content-Type: 'application/json',
  },
  postData: {
    type: 'account-membership',
    resourceId: 'Hopo4g34sLVdjEMBs2p19F',
    resourceType: 'account',
    recipientAlias: 'user@org.com',
    params: {
      role: 'cashier',
    },
  }
};
```

src/components/RequestCodePanel.astro
```diff
@@ -0,0 +1,23 @@
+---
+import { HTTPSnippet } from 'httpsnippet';
+const { title = 'Request', filename } = Astro.props;
+import { Code } from 'astro:components';
+import CodePanel from './CodePanel.astro';
+
+const file = await import(`../content/api/_code-snippets/dist/${filename}.json`);
```
tom-blake, line 7: @MeganSteenkamp Suggestion: Store generated HAR files outside of the `src` directory. Keeps things cleaner for local development

scripts/create-code-snippets.js
MeganSteenkamp: @tom-blake what did you have in mind for this? Validating what they provide like ensuring there is a path, method and headers?

MeganSteenkamp: If you don't provide something that is required your build will fail when we try to generate the HTTP snippet so I'm not sure what specific validation we need

src/components/RequestCodePanel.astro
MeganSteenkamp: @tom-blake I've moved this to a top level directory called `code-snippets` - let me know if you have a suggestion on where you were wanting the generated files to go 🎉

scripts/create-code-snippets.js
tom-blake: @MeganSteenkamp Great - that should be enough as long as we don't stray from the interface that the library expects. Does it fail with incorrect data types too?

You should definitely still add testing around our converter to make sure it is robust 👍

src/integrations/codeSnippets.js
```diff
@@ -0,0 +1,61 @@
+'use strict';
+
+import yaml from 'js-yaml';
+import fs from 'fs';
+import path from 'path';
+
+function transformYamlToHAR(data) {
```
tom-blake, line 7: @MeganSteenkamp Suggestion: Keep this code as a standalone module and wrap it in an Astro integration. Makes testing less convoluted

scripts/create-code-snippets.js
MeganSteenkamp: @tom-blake yes it does - here is an example where in my JS file I made the headers an array instead of an object.
image

src/content/api/_code-snippets/src/invitations-create.yaml
MeganSteenkamp: @tom-blake thanks I have updated this 🚀 
I had only thought about JSON and YAML and thought YAML is simpler but JS is even better.

src/components/RequestCodePanel.astro
```diff
@@ -0,0 +1,25 @@
...
+
+const file = await import(`../../code-snippets/${filename}.json`);
+const snippet = new HTTPSnippet(file);
+const request = snippet.requests[0];
+console.log({ request });
+const path = request.uriObj.pathname;
+const method = request.method;
+
+const options = { short: true, prettifyJson: true };
+const output = snippet.convert('shell', 'curl', options);
```
MeganSteenkamp, line 15: @alexBinns to generate another language you just need one more line to do another conversion.
An example would be:
`const output2 = snippet.convert('node', 'request', options);`

Here is the output:

image

src/content/api/_code-snippets/src/invitations-accept.js
```diff
@@ -0,0 +1,12 @@
+export default {
```
tom-blake, line 1: @MeganSteenkamp Suggestion: Remove the `src` subdirectory to simplify for devs

```
_code-snippets/
    dist/
    invitations-accept.js
    ...
```

package.json
```diff
@@ -29,8 +29,11 @@
     "eslint-plugin-standard": "^4.0.0",
     "eslint-plugin-vue": "^7.9.0",
     "flexsearch": "^0.7.31",
+    "fs": "^0.0.1-security",
     "glob": "^8.1.0",
     "gray-matter": "^4.0.3",
+    "httpsnippet": "^3.0.1",
+    "js-yaml": "^4.1.0",
```
tom-blake, line 36: @MeganSteenkamp Change: Remove `js-yaml`

src/utils/__tests__/transformToHarRequest.test.js
```diff
@@ -0,0 +1,79 @@
...
+        mimeType: 'application/json',
+        text: '{\"key\":\"value\"}',
+      }
+    };
+    const expectedString = JSON.stringify(expected, null, 2); // Pretty print JSON
+
+    expect(transformToHarRequest(data)).toBe(expectedString);
+  });
+
+  it('does not require headers', () => {
```
MeganSteenkamp, line 64: I'm not sure whether we should require headers or not. All of our requests in our docs tend to have an auth header.

src/utils/transformToHarRequest.js
```diff
@@ -0,0 +1,15 @@
+export function transformToHarRequest(data) {
+  const harObject = {
+    method: data.method,
+    url: new URL(data.path, 'https://service.centrapay.com/').href,
+    httpVersion: 'HTTP/2.0',
+    headers: data.headers && Object.entries(data.headers).map(([name, value]) => ({ name, value })),
```
tom-blake, line 6: @MeganSteenkamp Do we still have the issue where we can't include variables in our curl command? 

Can we do anything to solve it here?

src/utils/transformToHarRequest.js
```diff
@@ -0,0 +1,15 @@
+export function transformToHarRequest(data) {
```
tom-blake, line 1: @MeganSteenkamp Thoughts on creating a generic "code-snippet" module that wraps the entire snippet generation pipeline (our converter AND the httpsnippet converter)?

MeganSteenkamp: @tom-blake are you thinking that our `src/components/RequestCodePanel.astro` would call this module?

src/utils/transformToHarRequest.js
MeganSteenkamp: @tom-blake yeah unfortunately we can only generate with single quotes.
I can't find a solution to address this.

src/utils/transformToHarRequest.js
MeganSteenkamp: I'm not sure because at the moment the generator pipeline processes all the files whereas in `RequestCodePanel.astro` we just want to call the httpsnippet converter for one snippet.

MeganSteenkamp: I'm not sure because at the moment the converter runs for all files. This is kind of nice because we can clear out the dist directory. Whereas we only want to call `httpsnippet` for a specific file 🤔

src/utils/__tests__/transformToHarRequest.test.js
MeganSteenkamp: Resolution: Discussed this and have decided to leave as is seeing as we do have some public endpoints that could be documented with no headers.

src/utils/transformToHarRequest.js
MeganSteenkamp: Outcome: Discussed this and changed the Api Key variable to `<TOKEN>` which makes it clearer that you need to put your variable in.

src/utils/transformToHarRequest.js
MeganSteenkamp: Outcome: This has been refactored and now we don't need any intermediate files YAY 🎉

src/utils/__tests__/transformToHarRequest.test.js
```diff
@@ -0,0 +1,75 @@
+import transformToHarRequest from '../transformToHarRequest';
+
+describe('transformToHarRequest', () => {
+  it('transforms GET request', () => {
+    const data = {
+      method: 'GET',
```
tom-blake, line 6: @MeganSteenkamp We have `PUT` and `DELETE` APIs too :)

src/components/RequestCodePanel.astro
```diff
@@ -0,0 +1,24 @@
+---
+const { title = 'Request', filename } = Astro.props;
+import { Code } from 'astro:components';
+import CodePanel from './CodePanel.astro';
+import createRequestSnippet from '../utils/createRequestSnippet.js';
+
+const file = await import(`../content/api/_code-snippets/${filename}.js`);
+const snippet = createRequestSnippet(file.default);
```
tom-blake, line 8: @MeganSteenkamp How much more of this could be encapsulated within our module?

```js
// const snippet = centrapaySnippet.convert(filename, target, clientId);

const curlSnippet = centrapaySnippet.convert(filename, 'shell', 'curl');
```

MeganSteenkamp: @tom-blake I didn't do more because I think we just want the snippet object back.
The module could read the file though.

Having the snippet means we can get the path and the filename which we use in our object:
```
const request = snippet.requests[0];
const path = request.uriObj.pathname;
const method = request.method;
```

src/utils/__tests__/transformToHarRequest.test.js
MeganSteenkamp: @tom-blake do you think those need testing?
Maybe I should change the description of the tests as I think GET and POST cover all scenarios - i.e. post data and no post data.

tom-blake: @MeganSteenkamp Yep making the test generic is fine

Suggestion: Rename `postData` to `payload`. It's a more general name 👍

src/components/RequestCodePanel.astro
tom-blake: @MeganSteenkamp Thanks for the context 👍  I think our components shouldn't need to care about low-level httpsnippet details.

I see these new files as the declarative format of our API reference going forward - eventually everything will be in them. `path` and the `method` could be included in the data structure that our converter returns too:

```js
{
    title: 'Invitations',
    description: '...',
    path: '/api/invitations',
    method: 'POST',
    properties: [
        { name: '...', required: true, ..., },
    ],
    request: '<request string here>', // this is the httpsnippet output
    response: '<response string here>'
}
```

Thoughts?

MeganSteenkamp: @tom-blake I see what you're saying - I think this makes sense. 
I'll update this later today 🫡

MeganSteenkamp: @tom-blake I've updated this now 🫠

alexBinns (approved):

src/utils/Endpoint.js
```diff
@@ -0,0 +1,41 @@
...
+    this.method = props.method;
+    this.requests = props.requests;
+  }
+
+  static create(data) {
+    const harRequest = transformToHarRequest(data);
+
+    const snippet = new HTTPSnippet(harRequest);
+    const path = snippet.requests[0].uriObj.pathname;
+    const method = snippet.requests[0].method;
```
tom-blake, line 23: @MeganSteenkamp Suggestion: Use `path` and `method` from `data instead of relying on the HAR output

src/utils/Endpoint.js
```diff
@@ -0,0 +1,41 @@
...
+import { HTTPSnippet } from 'httpsnippet';
+
+const supportedClients = {
+  curl: {
+    targetId: 'shell',
+    options: { short: true, indent: ' ', prettifyJson: true },
+  },
+};
+
+class Endpoint {
```
tom-blake, line 11: @MeganSteenkamp Suggestion: Remove this class wrapper to reduce boilerplate. Just export a `create` function instead

src/utils/Endpoint.js
```diff
@@ -0,0 +1,41 @@
...
+  constructor(props) {
+    this.path = props.path;
+    this.method = props.method;
+    this.requests = props.requests;
+  }
+
+  static create(data) {
+    const harRequest = transformToHarRequest(data);
```
tom-blake, line 19: @MeganSteenkamp Suggestion: Remove this module and add the logic into this one instead.

Thoughts:
+ I'm not convinced this module is providing enough value to warrant the extra degree of separation. We're using snapshot tests for this module and we don't need to test the HAR library. Validation could be tested using files
+ We're exposing http snippet details in 2 places (this module and `snippet.convert below`). We should wrap it in a single place. I'm thinking something like `code: createSnippet()` on line 29 instead?

tom-blake (approved):

src/utils/Endpoint.js
MeganSteenkamp: @tom-blake I've updated this but I kept the creation of the snippet and the conversion to a string of code separate. The reasoning was that it doesn't make sense to create the same HAR Request repeatedly for each language you want to create a snippet for.

.gitignore
```diff
@@ -3,6 +3,7 @@
 
 # build output
 dist/
+code-snippets/
```
tom-blake, line 6: @MeganSteenkamp Change: Revert this 😄

src/utils/createEndpoint.js
```diff
@@ -0,0 +1,58 @@
...
+  const requiredFields = [ 'method', 'path'];
+  requiredFields.forEach(field => {
+    if (!(field in data)) {
+      throw new Error(`Request is missing required field ${field}`);
+    }
+  });
+}
+
+function createSnippet(data) {
+  validateData(data);
```
tom-blake, line 24: @MeganSteenkamp Suggestion: Perform validation before anything else (in the `createEndpoint` function)

src/utils/createEndpoint.js
```diff
@@ -0,0 +1,58 @@
...
+    harObject.postData = {
+      mimeType: 'application/json',
+      text: JSON.stringify(data.request.payload)
+    };
+  }
+  return new HTTPSnippet(harObject);
+}
+
+export default function createEndpoint(data) {
+  const snippet = createSnippet(data);
```
tom-blake, line 43: @MeganSteenkamp Suggestion: Call this `createSnippets` (plural). It would allow you to keep the snippet logic in one place and it would keep `createEndpoint` cleaner for future extensions

```js
const requests = createSnippets(data)
```

PR #891, opened 2023-11-22
Expandable Attributes

AndyClifford:
Attributes and Error sections are collapsed by default to save space. If the user is interested in an endpoint they can expand these sections to show the descriptions of each entry.

Also removed the default styling from the <button> html tag as per this [pull request](https://bitbucket.org/centrapay/quartz/pull-requests/60?atlOrigin=eyJpIjoiMWRiZjlmZjhkYmE3NDg0Mzk3NWI3ODZhZjczNGQyODQiLCJwIjoiYmItY2hhdHMtaW50ZWdyYXRpb24ifQ).

The first commit is the functional change. The second is just updating existing markdown to adhere to the new format.

**Ticket Link:** https://www.notion.so/centrapay/Lanterns-d784e8a5f9fa44328a5d0ae1fd85d37e?p=8ca71f65a68f4139b5960554cc6bbbff&pm=s

**Test Plan:**
- [ ] Got to docs.centrapay.com/api/accounts and observe the expandable sections work as expected.
- [ ] There should not be any double ups of headings for the same section

**Screenshots:**

- Mobile collapsed

image

- Mobile Expanded

image

- Desktop Collapsed

image

- Desktop Expanded

image

- Endpoint

image

- Endpoint with Errors

image

### DIFF
--- a/src/components/Error.astro
+++ b/src/components/Error.astro
@@ -4,12 +4,12 @@ import Badge from '../components/Badge.astro';
 const { code, message } = Astro.props;
 ---
 
-<li class="not-prose m-0 p-0 flex flex-col py-4 first:pt-0 space-y-2 type-content-primary">
+<li class="not-prose m-0 p-0 flex flex-col py-4 first:pt-0 type-content-primary">
   <div class="flex flex-row items-center space-x-3">
     <h4 class="font-mono" id={message}>{message}</h4>
     <Badge text={code}></Badge>
   </div>
-  <div data-expandable-slot class="hidden">
+  <div data-expandable-slot class="transition-opacity duration-500 collapsed">
     <slot />
   </div>
 </li>
--- a/src/components/Error.astro
+++ b/src/components/Error.astro
@@ -4,10 +4,12 @@ import Badge from '../components/Badge.astro';
 const { code, message } = Astro.props;
 ---
 
-<li class="not-prose m-0 p-0 flex flex-col py-4 space-y-2 type-content-primary">
+<li class="not-prose m-0 p-0 flex flex-col py-4 first:pt-0 space-y-2 type-content-primary">
   <div class="flex flex-row items-center space-x-3">
     <h4 class="font-mono" id={message}>{message}</h4>
     <Badge text={code}></Badge>
   </div>
-  <slot />
+  <div data-expandable-slot class="hidden">
+    <slot />
+  </div>
 </li>
--- a/src/components/Properties.astro
+++ b/src/components/Properties.astro
@@ -5,7 +5,7 @@ const { heading = 'Attributes' } = Astro.props;
 <div data-attributes class="my-6">
   <div class="flex justify-between items-center not-prose py-2">
     <h3 class="type-subtitle-1">{ heading }</h3>
-    <button data-expand-button class="text-content-tertiary type-overline">
+    <button data-expand-button class="text-content-tertiary type-overline hover:text-content-secondary">
       Expand all
     </button>
   </div>
@@ -28,12 +28,32 @@ const { heading = 'Attributes' } = Astro.props;
       button.textContent = button.innerText === 'EXPAND ALL' ? 'COLLAPSE ALL' : 'EXPAND ALL';
       const expandables = attributeList.querySelectorAll('[data-expandable-slot]');
       for (const expandable of expandables) {
-        if (expandable.classList.contains('hidden')) {
-          expandable.classList.remove('hidden');
+        if (expandable.classList.contains('collapsed')) {
+          expandable.classList.remove('collapsed');
+          expandable.classList.add('expanded');
         } else {
-          expandable.classList.add('hidden');
+          expandable.classList.remove('expanded');
+          expandable.classList.add('collapsed');
         }
       }
     });
   }
 </script>
+
+<style is:global>
+  .expanded {
+    @apply
+      pt-2
+      opacity-100
+      scale-y-100
+    ;
+  }
+
+  .collapsed {
+    @apply
+      h-0
+      opacity-0
+      scale-y-0
+    ;
+  }
+</style>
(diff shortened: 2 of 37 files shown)


src/components/Properties.astro
```diff
@@ -1,8 +1,49 @@
-<div class="my-6">
+---
+const { heading = 'Attributes' } = Astro.props;
+---
+
+<div data-attributes class="my-6">
+  <div class="flex justify-between items-center not-prose py-2">
+    <h3 class="scroll-mt-10 type-subtitle-1">{ heading }</h3>
```
MeganSteenkamp, line 7: Suggestion: Are these ever clickable titles? If not we probably don't need the scroll margin.

src/components/Properties.astro
```diff
@@ -1,8 +1,49 @@
...
+    <button data-expand-button class="text-content-tertiary type-overline">
+      Expand all
+    </button>
+  </div>
   <ul
+    data-attribute-list
     role="list"
-    class="m-0 list-none divide-y divide-zinc-900/5 p-0"
+    class="m-0 list-none border-y border-zinc-900/5 divide-y divide-zinc-900/5 p-0 pt-4"
```
MeganSteenkamp, line 15: Suggestion: Can the border colour be our colour token `border-outline-transparent`?

src/components/Properties.astro
```diff
@@ -1,8 +1,49 @@
-<div class="my-6">
+---
+const { heading = 'Attributes' } = Astro.props;
+---
+
+<div data-attributes class="my-6">
+  <div class="flex justify-between items-center not-prose py-2">
+    <h3 class="scroll-mt-10 type-subtitle-1">{ heading }</h3>
+    <button data-expand-button class="text-content-tertiary type-overline">
```
MeganSteenkamp, line 8: Change: Can we check what the hover states are for the button? I would expect there should be some.

src/components/Properties.astro
```diff
@@ -1,8 +1,49 @@
...
     role="list"
-    class="m-0 list-none divide-y divide-zinc-900/5 p-0"
+    class="m-0 list-none border-y border-zinc-900/5 divide-y divide-zinc-900/5 p-0 pt-4"
   >
     <slot />
   </ul>
 </div>
+
+<script>
+  function getElement(
```
MeganSteenkamp, line 22: Suggestion: Could this be pulled out into a helper function seeing as we have it in the code twice?

src/components/Properties.astro
AndyClifford: I thought a followup change might be to add ids to these fields so we can index them and include a search entry for them. Maybe its not a requirement though so will remove for now.

src/components/Properties.astro
AndyClifford: Thats even more transparent than we currently have

src/components/Properties.astro
AndyClifford: there aren't any in the design by the looks of it  - do you have a suggestion for what it could be?

src/components/Properties.astro
MeganSteenkamp: That's what's in the design 🤷‍♀️

src/components/Properties.astro
MeganSteenkamp: @AndyClifford ask the jesters to give you a design 😄

src/components/Properties.astro
```diff
@@ -1,8 +1,41 @@
...
+
+  for (const attributesSection of document.querySelectorAll('[data-attributes]')) {
+    const button = getElement('[data-expand-button]', HTMLButtonElement, attributesSection);
+    const attributeList = getElement('[data-attribute-list]', HTMLUListElement, attributesSection);
+    button.addEventListener('click', () => {
+      button.textContent = button.innerText === 'EXPAND ALL' ? 'COLLAPSE ALL' : 'EXPAND ALL';
+      const expandables = attributeList.querySelectorAll('[data-expandable-slot]');
+      for (const expandable of expandables) {
+        if (expandable.classList.contains('h-0')) {
+          expandable.classList.remove('h-0', 'opacity-0', 'scale-y-0');
```
MeganSteenkamp, line 32: @AndyClifford IMO this would be nicer to just add one class and you define the styles in Tailwind separately.

AndyClifford: Okay, just in one of the css files / a new file?

AndyClifford: Yeah I forgot to use those classes, I think just define it inline?

MeganSteenkamp: @AndyClifford I was thinking the `<style>` section on the Property component makes the most sense.

src/components/Error.astro
```diff
@@ -4,10 +4,30 @@ import Badge from '../components/Badge.astro';
...
     <Badge text={code}></Badge>
   </div>
-  <slot />
+  <div data-expandable-slot class="transition-all duration-500 collapsed">
+    <slot />
+  </div>
 </li>
+
+<style>
+  .expanded {
```
MeganSteenkamp, line 18: @AndyClifford maybe we should move this stying to the properties so that it isn't repeated. Sorry, I forgot about the errors section. 
You should be able to use global styling if you need: https://docs.astro.build/en/guides/styling/#global-styles

src/components/Properties.astro
```diff
@@ -1,8 +1,59 @@
-<div class="my-6">
+---
+const { heading = 'Attributes' } = Astro.props;
+---
+
+<div data-attributes class="my-6">
+  <div class="flex justify-between items-center not-prose py-2">
+    <h3 class="type-subtitle-1">{ heading }</h3>
+    <button data-expand-button class="text-content-tertiary type-overline hover:text-content-secondary">
```
AndyClifford, line 8: @MeganSteenkamp I added a hover state to the button - Jesters weren't keen on underline

MeganSteenkamp (approved):

PR #892, opened 2023-11-26
✨ Digital Assets: Generate code snippets

MeganSteenkamp:
Story: https://www.notion.so/centrapay/Use-httpSnippet-to-validate-and-generate-code-blocks-bb9ff3f377f644ccbc2030851af37d27?pvs=4

Test plan: Expect valid cURL requests in Digital Assets section.

Example endpoint with custom base URL and query string: http://docs-preview.cp42.click/guides/integrating-third-party-asset/#get-transaction-endpoint

### DIFF
--- a/src/content/api/_endpoints/asset-transfers-create-with-email.js
+++ b/src/content/api/_endpoints/asset-transfers-create-with-email.js
@@ -1,0 +1,14 @@
+export default {
+  method: 'POST',
+  path: '/asset-transfers',
+  request: {
+    headers: {
+      'X-Api-Key': '<TOKEN>',
+      'Content-Type': 'application/json',
+    },
+    payload: {
+      assetId: 'YGRo6TYYSxH3js7',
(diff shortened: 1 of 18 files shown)


src/content/api/_endpoints/asset-transfers-create-with-email.js
MeganSteenkamp, line 1: @tom-blake one potential issue I can see is that there are repetitions of the same endpoint with slightly varying payloads. I'm not sure if we need a better pattern for this - I guess we can see how much repetition we have one we migrate more pages.

MeganSteenkamp:
We are deferring figuring out a pattern for managing multiple payload variations for the same API endpoint.
Tech debt entry: https://www.notion.so/centrapay/Centrapay-Docs-Code-snippets-do-not-support-payload-variations-for-the-same-API-endpoint-d9e896bcc215413e9ef453101f183063

PR #893, opened 2023-11-27
Individual attributes collapsable

AndyClifford:
Added functionality for individual attributes to be collapsed by default and expand on click.

**Ticket:** https://www.notion.so/centrapay/Individual-attributes-expandable-6f0caabd385f47529007736ffad280bc?pvs=4

**Design:** https://www.figma.com/file/IgRlQWJTyGoZhlF4fVKD6B/Centrapay-Docs-(Tailwind)?type=design&node-id=1654-8412&mode=design&t=oTri02m3BIXILSXA-0

**Test Plan:**
- [ ] Go to docs.centrapay.com/api/accounts and use the collapsable sections. Make sure the expand/collapse functionality works as expected and the chevron is in the correct position on hover

**Screenshots:**

image

image

image

### DIFF
--- a/src/components/Error.astro
+++ b/src/components/Error.astro
@@ -1,15 +1,17 @@
 ---
 import Badge from '../components/Badge.astro';
+import Carat from './icons/Carat.vue';
 
 const { code, message } = Astro.props;
 ---
 
-<li class="not-prose m-0 p-0 flex flex-col py-4 first:pt-0 type-content-primary">
-  <div class="flex flex-row items-center space-x-3">
+<li data-attribute class="group not-prose m-0 p-0 flex flex-col py-4 first:pt-0 type-content-primary">
+  <button class="w-full relative flex flex-row items-center gap-x-3">
+    <Carat class="absolute -left-5 w-3 h-3 text-content-tertiary invisible" id="carat" data-carat />
     <h4 class="font-mono" id={message}>{message}</h4>
     <Badge text={code}></Badge>
-  </div>
-  <div data-expandable-slot class="transition-opacity duration-500 collapsed">
+  </button>
+  <div data-expandable class="transition-opacity duration-500 collapsed">
     <slot />
   </div>
 </li>
--- a/src/components/Properties.astro
+++ b/src/components/Properties.astro
@@ -10,7 +10,6 @@ const { heading = 'Attributes' } = Astro.props;
     </button>
   </div>
   <ul
-    data-attribute-list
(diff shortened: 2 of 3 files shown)


AndyClifford:
@MeganSteenkamp and I have been discussing whether this would be simpler as a Vue component for properties and errors to take better advantage of reactive state. It means we can use things like the disclosure component which is also better for accessibility.

If we use a Vue component we will need to pass in all the attributes as props instead of using slots. So something like:

```
<Properties properties=[
  { name="name", type="string", required=true, description="The name of the Account." },
  { name="type", type="string", required=true, description="Account type, must be either 'org' or 'individual'." },
  { name="owner", type="string", description="Id of user to add as member with 'account-owner' role." },
  { name="test", type="boolean", description="A flag indicating if the Account is for testing." },
  { name="region", type="string", description="The region that the Account will operate in. Required for 'org' Accounts, not allowed for 'individual' Accounts. Can be 'NZ', 'AU', or 'US'." },
]>
```

You might say that this is a step away from markdown. That's true but at the moment we are only utilising markdown for the description of individual properties so it isn't that bad in our opinion.

Vue components are easier to maintain as well as the dev team is more familiar with Vue as opposed to astro + raw JS.

MeganSteenkamp:
I think my preference is for the properties to just be a Vue component. I don't think the adding/ removing classes is scalable and you'll be able to get a more accessible UI.

src/components/Error.astro
```diff
@@ -1,15 +1,17 @@
...
+import Carat from './icons/Carat.vue';
 
 const { code, message } = Astro.props;
 ---
 
-<li class="not-prose m-0 p-0 flex flex-col py-4 first:pt-0 type-content-primary">
-  <div class="flex flex-row items-center space-x-3">
+<li data-attribute class="group not-prose m-0 p-0 flex flex-col py-4 first:pt-0 type-content-primary">
+  <button class="w-full relative flex flex-row items-center gap-x-3">
+    <Carat class="absolute -left-5 w-3 h-3 text-content-tertiary invisible" id="carat" data-carat />
```
MeganSteenkamp, line 10: One issue with the absolute positioning is that you don't get the cursor-pointer or behaviour of the button as it is outside the bounds of the button.

alexBinns:
What is happening with this PR @AndyClifford ?

alexBinns:
This has been open for a long time we can revisit this when we are back in the area

PR #896, opened 2023-11-27
WIP: Enable optional payload fields for code snippets

MeganSteenkamp:
Spike example for being able to create variations of the same API call in the guides.

### DIFF
--- a/src/content/api/_endpoints/payment-requests-create.js
+++ b/src/content/api/_endpoints/payment-requests-create.js
@@ -1,0 +1,46 @@
+export default {
+  method: 'POST',
+  path: '/api/payment-requests',
+  request: {
+    headers: {
+      'X-Api-Key': '<TOKEN>',
+      'Content-Type': 'application/json',
+    },
+    payload: {
+      required: {
+        configId: '5efbe2fb96c08357bb2b9242',
+        value: {
+          amount: '10000',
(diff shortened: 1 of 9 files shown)


src/content/api/_endpoints/payment-requests-create.js
MeganSteenkamp, line 1: @tom-blake this is WIP for what I want to meet with you about.
Yesterday we talked about potentially separating the API snippets from guide snippets entirely. I looked at the use case for payment requests and we had the following variations:
```
payment requests create
+ barcode
+ redirect url 
+ patronNotPresent, barcode, lineItems
+ conditionsEnabled, invoiceRef, purchaseOrderRef
+ conditionsEnabled, preAuth
+ conditionsEnabled, invoiceRef, purchaseOrderRef, patronNotPresent
```

If you have a separate file for each example, regardless of folder structure, how can you name all these variations? This made me think maybe we need a way to declare which optional fields you want present in your example.

MeganSteenkamp:
Recorded as tech debt: https://www.notion.so/centrapay/Centrapay-Docs-Code-snippets-do-not-support-payload-variations-for-the-same-API-endpoint-d9e896bcc215413e9ef453101f183063?pvs=4
Decision to only migrate API endpoints for now and figure out how to manage multiple request examples later.

PR #923, opened 2023-11-30
Snap to H2 headings

AndyClifford:
When hovering the H2 heading an anchor link appears allowing the consumer to snap to the heading. They can then easily copy the URL to return to the same spot in our docs later

**Test Plan:**
- [ ] Go to docs.centrapay.com/api/accounts and assert the anchor links snap as expected

**Screenys:**

image

image

image


src/components/H2.astro
```diff
@@ -1,8 +1,12 @@
 ---
+import AnchorLink from './icons/AnchorLink.vue';
 const { id } = Astro.props;
 ---
 
-<div class="not-prose">
+<div class="not-prose group flex items-center relative">
+  <a href=`#${id}` class="px-1.5 absolute right-full opacity-0 group-hover:opacity-100">
```
AndyClifford, line 7: TIL absolutely positioning an element 100% to the right of its parent will put it directly to the left of it. Having it directly to the left with no gap means the anchor link won't disappear when you try to hover over it

src/components/icons/AnchorLink.vue
```diff
@@ -0,0 +1,28 @@
...
+    height="16"
+    viewBox="0 0 16 16"
+    fill="currentColor"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
+      fill-rule="evenodd"
+      clip-rule="evenodd"
+      d="M4.5 5.75C3.90326 5.75 3.33097 5.98705 2.90901 6.40901C2.48705 6.83097 2.25 7.40326 2.25 8C2.25 8.59674 2.48705 9.16903 2.90901 9.59099C3.33097 10.0129 3.90326 10.25 4.5 10.25H6.27063C6.68484 10.25 7.02063 10.5858 7.02063 11C7.02063 11.4142 6.68484 11.75 6.27063 11.75H4.5C3.50544 11.75 2.55161 11.3549 1.84835 10.6517C1.14509 9.94839 0.75 8.99456 0.75 8C0.75 7.00544 1.14509 6.05161 1.84835 5.34835C2.55161 4.64509 3.50544 4.25 4.5 4.25H6.23156C6.64578 4.25 6.98156 4.58579 6.98156 5C6.98156 5.41421 6.64578 5.75 6.23156 5.75H4.5Z"
+      fill="#6B7280"
```
MeganSteenkamp, line 13: Suggestion: This should be `currentColor` and we pass one of our colour tokens in.

MeganSteenkamp:
@AndyClifford change: This currently only works for the API docs.

src/components/H2.astro
```diff
@@ -1,8 +1,12 @@
...
+    <AnchorLink class="w-4 h-4" />
+  </a>
```
MeganSteenkamp, line 9: Suggestion: You should be able to click the heading text and have the link snap.

PR #954, opened 2023-12-12
Get Payment Request Summary

AndyClifford:
**Ticket:** https://www.notion.so/centrapay/FE-Redirect-to-GiftStation-app-when-scanning-the-QR-code-78b3dfb0ea4b48ea9728b63a93b773da?pvs=4

**Test Plan:**
- [ ] Go to docs.centrapay.com/api/payment-requests#get-payment-request-summary
- [ ] Assert the description, request and response are as expected.

### DIFF
--- a/src/__tests__/__snapshots__/links.integration.test.js.snap
+++ b/src/__tests__/__snapshots__/links.integration.test.js.snap
@@ -218,6 +218,7 @@ exports[`the build should not break any links 1`] = `
   "/api/payment-requests#get-a-payment-request",
   "/api/payment-requests#get-a-payment-request-by-short-code",
   "/api/payment-requests#get-a-payment-request-linked-to-a-patron-code",
+  "/api/payment-requests#get-payment-request-summary",
   "/api/payment-requests#line-item-model",
   "/api/payment-requests#list-payment-activities-for-a-merchant",
   "/api/payment-requests#list-payment-activities-for-a-payment-request",
--- a/src/content/api/payment-requests.mdx
+++ b/src/content/api/payment-requests.mdx
@@ -573,6 +573,17 @@ Payment Activities are created when a Payment Request has been `created`, `paid`
 ---
 
 <Endpoint
+  path="/api/payment-requests/{paymentRequestId}/summary"
+  filename="payment-requests-get-summary"
+>
+  ## Get Payment Request Summary
+
+  This endpoint allows you to retrieve the summary of a Payment Request while the status is `new`.
+</Endpoint>
+
+---
+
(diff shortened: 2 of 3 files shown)


src/__tests__/__snapshots__/links.integration.test.js.snap
```diff
@@ -24,16 +24,6 @@ exports[`the build should not break any links 1`] = `
...
-  "/api/asset-activities",
-  "/api/asset-activities#asset-activity-model",
-  "/api/asset-activities#asset-created-activity-model",
-  "/api/asset-activities#asset-received-activity-model",
-  "/api/asset-activities#asset-sent-activity-model",
-  "/api/asset-activities#asset-transaction-activity-model",
-  "/api/asset-activities#list-asset-activities",
-  "/api/asset-activities#list-asset-transactions",
-  "/api/asset-activities#token-redeemed-activity-model",
-  "/api/asset-activities#token-refunded-activity-model",
```
Neeraj-Ghodla, line 36: why are all these lines removed?

AndyClifford: Oops will fix this now

src/content/api/payment-requests.mdx
```diff
@@ -572,6 +572,17 @@ Payment Activities are created when a Payment Request has been `created`, `paid`
 
 ---
 
+<Endpoint
+  path="/api/payment-requests/{paymentRequestId}/summary"
+  filename="payment-requests-get-summary"
+>
+  ## Get Payment Request Summary
+
+  This endpoint allows you to retrieve the summary of a Payment Request while the status is `new`.
```
Neeraj-Ghodla, line 581: I think you should also say somewhere in the description that you don't need to be authenticated to call this endpoint.

AndyClifford: Yeah was wondering about that - I'm not sure why the asset summary endpoints don't say anything. I guess the example request shows that you don't need an auth token but it still isn't that clear.

Do you think a `public` badge or something similar would be a good option?

Neeraj-Ghodla: Yes, a public badge seems like a good idea.

AndyClifford: I've added a followup task for this as there is no design for this yet https://www.notion.so/centrapay/Add-public-badge-to-all-public-routes-in-Centrapay-docs-69c09d32b8154530a6974a66859f2497?pvs=4

Neeraj-Ghodla (approved):

PR #957, opened 2023-12-19
Add public badge for get payment request summary endpoint

alexBinns:
This PR adds the public badge to get payment request summary endpoint

### DIFF
--- a/src/content/api/payment-requests.mdx
+++ b/src/content/api/payment-requests.mdx
@@ -9,6 +9,7 @@ import Properties from '../../components/Properties.astro';
 import Property from '../../components/Property.astro';
 import Error from '../../components/Error.astro';
 import Endpoint from '../../components/Endpoint.astro';
+import Badge from '../../components/Badge.astro';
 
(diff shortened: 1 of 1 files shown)


Neeraj-Ghodla (approved):

AndyClifford:
We should add this badge to the assets/asset transfer public summary endpoints

src/content/api/payment-requests.mdx
```diff
@@ -576,7 +577,7 @@ Payment Activities are created when a Payment Request has been `created`, `paid`
   path="/api/payment-requests/{paymentRequestId}/summary"
   filename="payment-requests-get-summary"
 >
-  ## Get Payment Request Summary
+  ## Get Payment Request Summary <Badge text="public" />
```
AndyClifford, line 580: Was the design to use the default color?

alexBinns: Yea it was but @tom-blake didn't like adding the public badge as all the endpoints are public. There is a discussion in the design channel about alternative designs

alexBinns:
This solution is not ideal and there is some more thought that needs to go into how we display these.

PR #963, opened 2024-02-08
Update merchant create and update routes to include the refund config…

daallen303:
Add refund config property to the merchant create and update docs

### DIFF
--- a/src/content/api/merchants.mdx
+++ b/src/content/api/merchants.mdx
@@ -58,6 +58,10 @@ which define the payment methods available for a Payment Request.
     Merchant [Settlement Config](#settlement-config-model).
   </Property>
 
+  <Property name="refundConfig" type="object">
+    Merchant [Refund Config](#refund-config-model).
+  </Property>
+
   <Property name="location" type="location" experimental>
     Physical Location of Merchant. It is highly recommended that you provide this otherwise, users won't be able to find you with our [Merchant Search](#search-merchants) API if they perform a origin + distance query.
   </Property>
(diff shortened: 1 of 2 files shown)


src/content/api/merchants.mdx
```diff
@@ -183,6 +196,10 @@ which define the payment methods available for a Payment Request.
       Merchant [Settlement Config](#settlement-config-model).
     </Property>
 
+    <Property name="refundConfig" type="object">
```
laurencemcknight, line 199: Should we add the experimental tag as this will likely be a model that changes over time?

src/content/api/merchants.mdx
```diff
@@ -58,6 +58,10 @@ which define the payment methods available for a Payment Request.
     Merchant [Settlement Config](#settlement-config-model).
   </Property>
 
+  <Property name="refundConfig" type="object">
```
laurencemcknight, line 61: Should we add the experimental tag as this will likely be a model that changes over time?

daallen303: If it changes over time that's fine as long as we don't remove the bank account id field or break the API contract we setup. I think we over use the experimental flag.

src/content/api/merchants.mdx
```diff
@@ -106,6 +110,15 @@ which define the payment methods available for a Payment Request.
 
 ---
 
+## Refund Config Model
+
+
+<Properties>
+  <Property name="bankAccountId" type="string">
+    The id of the bank account that will be used when processing refunds. This must belong to the account
```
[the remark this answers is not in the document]
daallen303: I guess only problem there is the bank account actually belongs to the account and not the merchant.

MeganSteenkamp (approved):

laurencemcknight (approved):

PR #982, opened 2024-03-14
Remove line about archived assets not being returned from listAssets API

Ray2126:
The listAssets endpoint is being changed to return all assets.


Ray2126:
We have decided not to return "archived" assets.

PR #995, opened 2024-04-05
Add content-type for the example code

subarashii-huch09:
Currently, when we copy the example curl from the `Create Promotion API` from the public doc, it returns a status code **400 - Bad Request**.

To fix this,   `-H 'content-type: application/json' \` is added 

The task story is [here](https://www.notion.so/centrapay/Minor-Public-Doc-Fix-Create-Promotion-API-3a6045e5455e464d98fda1ad342e7376?pvs=4)

The public doc can be previewed from [here](http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/api/promotions/)

## Test Plan
- Directly copy the example curl of the `Create Promotion API` from the public doc.
- Call the API, it should succeed and the error response of 400 Bad Request should not return

### DIFF
--- a/src/content/api/_endpoints/bank-accounts-archive.js
+++ b/src/content/api/_endpoints/bank-accounts-archive.js
@@ -4,7 +4,6 @@ export default {
   request: {
     headers: {
       'X-Api-Key': '<TOKEN>',
-      'Content-Type': 'application/json',
     }
   },
   response: {
--- a/src/content/api/_endpoints/promotions-create.js
+++ b/src/content/api/_endpoints/promotions-create.js
@@ -4,6 +4,7 @@ export default {
   request: {
     headers: {
       'X-Api-Key': '<TOKEN>',
+      'Content-Type': 'application/json',
     },
     payload: {
       name: 'Spend a Buck',


seliwe (approved):

src/content/api/_endpoints/promotions-create.js
```diff
@@ -4,6 +4,7 @@ export default {
   request: {
     headers: {
       'X-Api-Key': '<TOKEN>',
+      'Content-Type': 'application/json',
```
tom-blake, line 7: @subarashii-huch09 Suggestion: Check if any of our other APIs have this issue. Fix any others too 😄

subarashii-huch09: @tom-blake  Yep, great point! 🙂 I'll check other APIs too. 🫡

src/content/api/_endpoints/bank-accounts-archive.js
```diff
@@ -4,7 +4,6 @@ export default {
   request: {
     headers: {
       'X-Api-Key': '<TOKEN>',
-      'Content-Type': 'application/json',
     }
```
subarashii-huch09, line 7: @seliwe @tom-blake
I've gone through our APIs to ensure 'Content-Type': 'application/json' is included in relevant Post APIs.
The only thing that I noticed is this `Archive Bank Account` API.  It does not have Request body but it has 'Content-Type': 'application/json' included in the curl, so I've removed it.

seliwe (approved):

Ray2126 (approved):

PR #1005, opened 2024-05-07
Add conditionsDescription to the Promotion model

Ray2126:


Ray2126:
Dev Preview: http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/api/promotions/

Ray2126:
Should we only allow conditionsDescription to be provided if conditions is provided?

subarashii-huch09 (approved):
LGTM : )

seliwe:
> Should we only allow `conditionsDescription` to be provided if `conditions` is provided?

I left a comment on the Rhea PR that we might not want to lock down this description to being about conditions, e.g. it could be used to describe awards. Description fields tend to get abused for various purposes in the long term, so I'd name it generically and have it always optional.

seliwe:
> > Should we only allow `conditionsDescription` to be provided if `conditions` is provided?
> 
> I left a comment on the Rhea PR that we might not want to lock down this description to being about conditions, e.g. it could be used to describe awards. Description fields tend to get abused for various purposes in the long term, so I'd name it generically and have it always optional.

As discussed we might want additional descriptions in future e.g. rewardsDescription so I'll approve and leave you to resolve with designers

seliwe (approved):

Ray2126:
I've caught up with Junjie and he thinks this description will only ever be describing conditions. We don't think rewards will ever need descriptions. There might be terms and conditions in the future but this should be a different field.

PR #1014, opened 2024-05-27
Remove refund config from docs

daallen303:
Story: https://www.notion.so/centrapay/BE-Remove-refundConfig-f80a5329c0f5406bac9b141fdb9939ac?pvs=4

We are no longer going to be using the merchant refund config and will be removing it from the front and backend.


MeganSteenkamp:
@daallen303 Change: We should also remove the refund config model that was added here https://github.com/centrapay/centrapay-docs/pull/963 🤝

daallen303:
> @daallen303 Change: We should also remove the refund config model that was added here #963 🤝

@MeganSteenkamp good catch, done. :)

MeganSteenkamp (approved):

PR #1020, opened 2024-06-24
add fields for discount collections

seliwe:
These changes allow the creation of Collections of type discount, which will be used for Quartz Points. The monetary value of `maxValue` defines the amount that each point will discount from a payment. `tokenExpiresAt` allows for providing a hard end date to a token campaign.

http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/api/tokens/#create-token-collection


subarashii-huch09 (approved):
LGTM :)

Ray2126 (approved):

PR #1028, opened 2024-07-16
Fix business

AndyClifford:
Accidentally added to the wrong endpoint. Also added the same error to set business onboarding status


alexBinns (approved):

alexBinns:
@AndyClifford can we get this merged

PR #1034, opened 2024-07-24
Document type field for create merchant API

laurencemcknight:
We have added a `type` field to the create merchant API. This allows users to specify whether the merchant is a physical store or an online ecommerce merchant. The change documents the new field.

### DIFF
--- a/src/content/api/merchants.mdx
+++ b/src/content/api/merchants.mdx
@@ -186,6 +186,10 @@ which define the payment methods available for a Payment Request.
     <Property name="location" type="location">
       Physical Location of Merchant. It is highly recommended that you provide this otherwise, users won't be able to find you with our [Merchant Search](#search-merchants) API if they perform a origin + distance query.
     </Property>
+
+    <Property name="type" type="string">
+      Valid values: `physical` or `e-commerce`.
+    </Property>
   </Properties>
 
   <Properties heading="Errors">
--- a/src/content/api/merchants.mdx
+++ b/src/content/api/merchants.mdx
@@ -62,6 +62,10 @@ which define the payment methods available for a Payment Request.
     Physical Location of Merchant. It is highly recommended that you provide this otherwise, users won't be able to find you with our [Merchant Search](#search-merchants) API if they perform a origin + distance query.
   </Property>
 
+  <Property name="type" type="string">
+    Valid values: `physical` or `e-commerce`. Used for reporting purposes.
+  </Property>
+
   <Property name="onboardingStatus" type="string">
     The onboarding status of the Merchant. See [Onboarding Statuses](#onboarding-statuses) for possible values.
   </Property>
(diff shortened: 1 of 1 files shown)


src/content/api/merchants.mdx
```diff
@@ -186,6 +186,10 @@ which define the payment methods available for a Payment Request.
     <Property name="location" type="location">
       Physical Location of Merchant. It is highly recommended that you provide this otherwise, users won't be able to find you with our [Merchant Search](#search-merchants) API if they perform a origin + distance query.
     </Property>
+
+    <Property name="type" type="string">
+      Valid values: `physical` or `e-commerce`.
```
alexBinns, line 191: Do we need to add more information here as to what this value is used for?

laurencemcknight: It is used for reporting purposes. I'd ideally like to avoid including low level explanations such as that in our public docs. Do you agree?

alexBinns: I just think if we are expecting the integrator to make a decision then we should provide them enough info to make a decision

alexBinns: You probably need to add to the merchant model as well. Maybe that is the place to add more information.

src/content/api/merchants.mdx
```diff
@@ -62,6 +62,10 @@ which define the payment methods available for a Payment Request.
     Physical Location of Merchant. It is highly recommended that you provide this otherwise, users won't be able to find you with our [Merchant Search](#search-merchants) API if they perform a origin + distance query.
   </Property>
 
+  <Property name="type" type="string">
+    Valid values: `physical` or `e-commerce`. Used for reporting purposes.
+  </Property>
+
```
daallen303, line 68: I think you need to explain that location can't be provided for an e-commerce merchant. I think that would probably go in the explanation for the location above. I think also explaining that e-commerce merchants will not show up on the merchant search would be important.

daallen303: I don't think there is a need to mention reporting, maybe you could just say something like type should be physical for any merchant that has a physical location. If a merchant is going to be used for e-commerce payments the type should be 'e-commerce'.

laurencemcknight: This has been added to the model. See the code above.

alexBinns (approved):
LGTM

src/content/api/merchants.mdx
```diff
@@ -58,8 +58,12 @@ which define the payment methods available for a Payment Request.
     Merchant [Settlement Config](#settlement-config-model).
   </Property>
 
+  <Property name="type" type="string">
+    Valid values: `physical` or `e-commerce`. Merchants with a physical store should provide type `physical`. Online only merchants should provide type `e-commerce`. E-commerce merchants will not be displayed in the store finder.
```
daallen303, line 62: Do you think we should specify if type isn't provided that it will default to 'physical'?

laurencemcknight: That behaviour has changed, it will no longer default to `physical`.

PR #1039, opened 2024-07-29
Document new card number field in batch schema

laurencemcknight:
Updating the file schema for Farmlands card batch loading. This will allow us to ingest card numbers for future use in the payment flow.

### DIFF
--- a/src/content/api/batch-types/farmlands-external-asset.mdx
+++ b/src/content/api/batch-types/farmlands-external-asset.mdx
@@ -86,7 +86,7 @@ Exported Farmlands Credit Card information used for importing and updating of a 
   </Property>
 
   <Property name="number" type="bignumber">
-    Unique 16 digit field used in conjunction with `expiry` to pay by [Patron Code](/api/patron-codes).
+    Unique 16 digit identifier for the card that can be used in conjunction with `expiry` for eCommerce payments.
   </Property>
 
   <Property name="status" type="string">
--- a/src/content/api/batch-types/farmlands-external-asset.mdx
+++ b/src/content/api/batch-types/farmlands-external-asset.mdx
@@ -85,7 +85,7 @@ Exported Farmlands Credit Card information used for importing and updating of a 
     Unique 9 digit field to display in barcode format to pay by [Patron Code](/api/patron-codes).
   </Property>
 
-  <Property name="number" type="bignumber">
+  <Property name="number" type="PAN">
     Unique 16 digit identifier for the card that can be used in conjunction with `expiry` for eCommerce payments.
   </Property>
 
--- a/src/content/api/batch-types/farmlands-external-asset.mdx
+++ b/src/content/api/batch-types/farmlands-external-asset.mdx
@@ -86,7 +86,7 @@ Exported Farmlands Credit Card information used for importing and updating of a 
   </Property>
 
   <Property name="number" type="PAN">
-    Unique 16 digit identifier for the card that can be used in conjunction with `expiry` for eCommerce payments.
+    16 digit primary account number of the Card.
   </Property>
 
   <Property name="status" type="string">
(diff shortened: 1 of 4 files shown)


src/content/api/batch-types/farmlands-external-asset.mdx
```diff
@@ -85,6 +85,10 @@ Exported Farmlands Credit Card information used for importing and updating of a
     Unique 9 digit field to display in barcode format to pay by [Patron Code](/api/patron-codes).
   </Property>
 
+  <Property name="number" type="bignumber">
```
daallen303, line 88: I think potentially "primary account number" or PAN might be a more specific way of naming this field. Could you ask Neza or a contact at Farmlands if this would be correct? https://en.wikipedia.org/wiki/Payment_card_number

src/content/api/batch-types/farmlands-external-asset.mdx
```diff
@@ -85,6 +85,10 @@ Exported Farmlands Credit Card information used for importing and updating of a
...
+    Unique 16 digit field used in conjunction with `expiry` to pay by [Patron Code](/api/patron-codes).
```
daallen303, line 89: What do you think about 'Unique 16 digit identifier for the card that can be used for eCommerce payments'? pay by patron code makes me think of quick pay or the scan barcode payment flow and this feels separate even though implementation wise patron code might be the correct way it is modeled.

daallen303 (approved):

daallen303:
I think we will need another field that represents the day/month expiry present on the card. We could derive it from the expires at potentially but it might make sense for that to be a separate field that's sent.

src/content/api/batch-types/farmlands-external-asset.mdx
laurencemcknight: I like it. I'm going to keep in the mention that it must be used in conjunction with Expiry, I think that's important info and isn't related to quick pay.

laurencemcknight:
> I think we will need another field that represents the month/year expiry present on the card. We could derive it from the expires at potentially but it might make sense for that to be a separate field that's sent.

Could you explain why you prefer a seperate field? I imagine parsing the expiresAt into a month/year format would be quite straightforward.

daallen303:
> > I think we will need another field that represents the month/year expiry present on the card. We could derive it from the expires at potentially but it might make sense for that to be a separate field that's sent.
> 
> Could you explain why you prefer a seperate field? I imagine parsing the expiresAt into a month/year format would be quite straightforward.

Just feels like there could be an edge case that allows for bugs. There is a specific expiry that is on the physical card, that is not equal to the expiresAt that Farmlands provides us. It might cause an issue in the future if we assume we can just parse the expiresAt that is provided. This should be discussed with Farmlands and they should be able to give us an answer pretty quick.

MeganSteenkamp (changes requested):
Suggested action:
1. Review batch upload process.
2. Check whether card expiry and expiry timestamp mismatch is a possibility.

laurencemcknight:
> Suggested action:
> 
> 1. Review batch upload process.
> 2. Check whether card expiry and expiry timestamp mismatch is a possibility.

1. Batch upload process is unchanged
2. Mismatch is not a possibility. We therefore do not need to ingest shorthand expiry.

MeganSteenkamp (approved):

src/content/api/batch-types/farmlands-external-asset.mdx
```diff
@@ -85,6 +85,10 @@ Exported Farmlands Credit Card information used for importing and updating of a
     Unique 9 digit field to display in barcode format to pay by [Patron Code](/api/patron-codes).
   </Property>
 
+  <Property name="number" type="PAN">
+    Unique 16 digit identifier for the card that can be used in conjunction with `expiry` for eCommerce payments.
```
MeganSteenkamp, line 89: Suggestion: Simplify the description to say "16 digit primary account number of the card".
The PAN alone isn't unique. This is also decoupled from e-commerce so don't need to mention that implementation.

PR #1044, opened 2024-08-16
✨ Add ecommerce integration guide

MeganSteenkamp:
Update ecommerce guide to detail integration with Centrapay Commerce SDK.
- Add sequence diagram for ecommerce process.
- Add step by step integration instructions.
- Add sample integration code.

Style changes:
- Allow code blocks to have no max height if specified.

Test plan:
- Expect existing code blocks to have fixed max height.
- Expect ecommerce page to be updated per the preview: http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/guides/ecommerce-website/


src/content/guides/ecommerce-website.mdx
```diff
@@ -5,27 +5,125 @@ nav:
...
+        async onClick() {
+          // Create Payment Request
+          // Return Payment Request ID
+        },
+        async onComplete(data) {
+          if (data.paymmentRequest.status === 'new') {
+            // Void Payment Request
+          }
+        }
+      });
```
Neeraj-Ghodla, line 97: Did you consider separating this into two functions i.e. `onSuccess` and `onFailure`, where the `onSuccess` callback gets called on a successfull payment and `onFailure` callback gets called for unsuccessfull payments.

MeganSteenkamp: @Neeraj-Ghodla we could do that but it seems easier to implement one function. What benefit do your think that would give?

Neeraj-Ghodla: I though by separating the integrator don't need to worry about the status of the payment request when those callbacks are called, but I think even with the functions spearated you'd still need to check the status in the `onFailure` case because you can't void expired requests.

MeganSteenkamp: @Neeraj-Ghodla yes I think they would still need to check so will leave this as is. This page is experimental so we can always change it later if we see a need to.

laurencemcknight (approved):

PR #1068, opened 2024-09-30
🔥 Remove barcode required when preAuth is true

MeganSteenkamp:
This is not correct as we can now create payment requests with preAuth and no quick pay.

Test plan: Expect this line not to be present in docs.

### DIFF
--- a/src/content/api/payment-requests.mdx
+++ b/src/content/api/payment-requests.mdx
@@ -81,7 +81,7 @@ Centrapay Payment Requests are serviced via two sets of endpoints; the â€œnextâ€
     The id of a [Patron Code](/api/patron-codes) the Payment Request is attached to.
   </Property>
   <Property name="barcode" type="string">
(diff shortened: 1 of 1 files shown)


daallen303 (approved):

src/content/api/payment-requests.mdx
```diff
@@ -453,7 +453,7 @@ Payment Activities are created when a Payment Request has been `created`, `paid`
       The canonical value of the Payment Request. Must be less than 100000000 and positive.
     </Property>
     <Property name="barcode" type="string">
-      [Scanned Code](/api/scanned-codes) used to create the Payment Request. Required when `preAuth` is `true`.
+      [Scanned Code](/api/scanned-codes) used to create the Payment Request.
```
daallen303, line 456: Change: it is still required for quick pay I believe so I think it would be good to mention that.

MeganSteenkamp: @daallen303 Have updated this to reference the quick pay flow.
image

PR #1071, opened 2024-10-07
add ceil option to token expires after model

bpal938:
This change was recently made as a part of https://www.notion.so/centrapay/Payap-points-expiry-10ea9ab17e80809697addec8e78fa080

test plan: 
- go to https://docs.centrapay.com/api/tokens/#token-expires-after-model and expect the new ceil field to be visible

image

### DIFF
--- a/src/content/api/tokens.mdx
+++ b/src/content/api/tokens.mdx
@@ -91,12 +91,17 @@ A [Redemption Condition](#redemption-condition-model) is created for each [Merch
 
 
 <Properties>
-  <Property name="period" type="string">
+  <Property name="period" type="string" required>
     Supported values are `hour`, `day`, `week`, `month` and `year`.
   </Property>
 
-  <Property name="duration" type="string">
+  <Property name="duration" type="string" required>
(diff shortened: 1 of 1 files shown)


Neeraj-Ghodla (approved):

src/content/api/tokens.mdx
```diff
@@ -98,6 +98,11 @@ A [Redemption Condition](#redemption-condition-model) is created for each [Merch
   <Property name="duration" type="string">
     Number of `period` until token expiration.
   </Property>
+
+  <Property name="ceil" type="string">
+    Interval to round token expiry up to.
+    Supported values are `hour`, `day`, `week`, `month` and `year`.
+  </Property>
```
Neeraj-Ghodla, line 105: I think this field should also have an optional flag on it.

bpal938: There are no optional flags that we use in docs but I can add required to the other two properties

Wulingoh (approved):

PR #1080, opened 2024-11-12
⏰ [10% time]: ♻️ Make astro header component

MeganSteenkamp:
Changes:
1. Change header component to use Astro instead of Vue.
2. Refactor existing components to make header easier to understand.

Test plan: Expect header bar functionality to be unchanged. Regression test:
1. Click Centrapay logo, redirected home.
2. Click Docs, redirected home.
3. Click search, can enter term and click link.
4. Click '/' to open command palette, can then press escape to close.
5. Mobile: Can open mobile menu and click link.

Preview site: http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/


src/components/Header.astro
```diff
@@ -0,0 +1,64 @@
...
+      <div class="flex items-center">
+        <SiteSearch client:load />
+        <MobileMenu client:load data-mobile-menu path={path} navigation={navigation} />
+      </div>
+    </div>
+  </header>
+</centrapay-header>
+
+<script>
+  class CentrapayHeader extends HTMLElement {
```
MeganSteenkamp, line 40: See docs: https://docs.astro.build/en/guides/client-side-scripts/#web-components-with-custom-elements

src/components/MobileMenu.vue
MeganSteenkamp, line 1: This is existing logic, have just changed the component abstraction.
Note that there is an existing issue; when the mobile menu is open the buttons along the header are not able to be clicked. I think this may be due to using the HeadlessUI Dialog component but seeing as we plan to move this to Astro it's not worth changing right now.

src/components/SiteSearch.vue
MeganSteenkamp, line 1: Existing logic abstracted out of header.

src/components/Header.astro
```diff
@@ -0,0 +1,64 @@
...
+        return;
+      }
+
+      openBtn.addEventListener('click', () => {
+        openBtn.style.display = 'none';
+        closeBtn.style.display = 'flex';
+        mobileMenu.style.display = 'relative';
+      });
+
+      closeBtn.addEventListener('click', () => {
```
tom-blake, line 55: @MeganSteenkamp The mobile menu is broken on the preview site

1. Switch to mobile view
2. Open sidebar
3. Close sidebar - this doesn't work
4. Switching back to desktop view keeps the mobile menu open too

image

MeganSteenkamp: @tom-blake this is already broken, see this comment: https://github.com/centrapay/centrapay-docs/pull/1080#discussion_r1837538743

The only difference is right now if you go back to desktop in the current site it does close. I don't see this as a user issue as you are normally on mobile or desktop, not switching between. I think we should fix this separately.

tom-blake (approved):

Ray2126:
What's the reason for replacing the Header Vue component with an Astro component? This will make it harder to keep all of our frontends consistent when we eventually add a component library

MeganSteenkamp:
@Ray2126 the goal for the docs site is to not use Vue at all; only Astro. 
This will improve the performance and rendering so it seems like this will be fully separate from our component library.

PR #1083, opened 2024-11-14
Upgrade to Node 22

tom-blake:
Node 18 is in maintenance mode. Node 22 is the currently active LTS version.

https://nodejs.org/en/about/previous-releases

Test plan:
- Confirm site runs successfully locally. Both with and without docker compose
- Confirm CI pipelines pass


MeganSteenkamp:
@tom-blake was this fine with httpsnippet?

MeganSteenkamp (approved):

PR #1087, opened 2024-11-15
Convert transaction reporting guide to markdoc

tom-blake:
- Convert transaction reporting guide to markdoc
- Add new Astro components for API links and informational notes to avoid affecting existing MDX pages.
- Add markdoc config to repository
- Remove obselete docker compose version to prevent warning locally

Test plan:
- Confirm transaction reporting guide has no visual changes
- Confirm API link correctly navigates to the right URL


src/content/guides/transaction-reporting.mdoc
```diff
@@ -8,8 +8,10 @@ nav:
```
tom-blake, line 8: New vs old

image

src/components/ApiLink.astro
```diff
@@ -0,0 +1,18 @@
...
+---
+<a
+  href={href}
+  target="_self"
+  class="group inline-flex justify-center items-center text-interactive-secondary hover:text-interactive-secondary-hover"
+>
+  <span class="underline">
+    {title}
+  </span>
+  <span>&nbsp;</span>
```
MeganSteenkamp, line 12: Question: Why do you need this additional span?

src/components/icons/InfoIcon.astro
MeganSteenkamp, line 1: @tom-blake Change: The icon looks slightly different, I think this is because the classes aren't passing through.
To fix it you need the following:
```
---
const { class: className } = Astro.props;
---

# add this to the SVG
class={className}
```

MeganSteenkamp (approved):

src/components/ApiLink.astro
tom-blake: @MeganSteenkamp It puts a space between the API badge and the link text. I've lifted and shifted the original existing `ProseA` component here - worth chatting about whether we should consider ditching the API badge in a follow up 👍

src/components/icons/InfoIcon.astro
tom-blake: Great spot - updated

PR #1089, opened 2024-11-17
Use expressive code to render code blocks

tom-blake:
Use expressive code to render code blocks instead of maintaining our own custom component.

Key notes:
- [ADR here](https://www.notion.so/centrapay/Use-expressive-code-to-render-code-blocks-in-our-docs-site-141a9ab17e8080c894aafa382f95e2c4)
- I've chosen to use the material-theme-palenight. [Full available theme list here](https://expressive-code.com/guides/themes/)
- Code blocks no longer include custom method and path props.
- MDX file changes only include moving the code block title inline with the markdown fence backticks.

Test plan:
- Confirm that code blocks render correctly for all pages and API endpoints when running the dev server and when the site is being built statically


ec.config.mjs
```diff
@@ -0,0 +1,8 @@
+import { defineEcConfig } from 'astro-expressive-code';
+
+export default defineEcConfig({
+  themes: ['material-theme-palenight'],
```
tom-blake, line 4: image

tom-blake: [Dev site is deployed here](http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/)

MeganSteenkamp: @tom-blake is there a way to set a max height? 
Some of the code blocks are really long - here is an example: http://centrapay-docs.dev.s3-website-ap-southeast-1.amazonaws.com/api/payment-requests/#list-payment-activities-for-a-merchant

MeganSteenkamp (approved):
Looks awesome 💪

PR #1090, opened 2024-11-17
Convert patron not present guide to markdoc

tom-blake:
Add new CodeFence component to allow markdoc to render code fences using a custom component.

The CodePanel component used by the MDX integration will be removed once we've migrated all guides to markdoc.

Test plan:
- Confirm no visual regressions on the patron not present guide
- Confirm code fences on MDX files are being rendered correctly


src/content/guides/patron-not-present.mdx
```diff
@@ -1,39 +0,0 @@
----
```
tom-blake, line 1: The changes to this file are called out below

src/content/guides/patron-not-present.mdoc
```diff
@@ -0,0 +1,35 @@
+---
```
tom-blake, line 1: Git is representing the changes as an insert / delete so I've called out the changes here
- File extension updated to `mdoc`
- We no longer need to explicitly import the custom fence component
- Code fences no longer need to be wrapped in a custom component
- Custom attributes like the code fence title need to be wrapped with markdoc tags

MeganSteenkamp (approved):

PR #1101, opened 2024-11-22
[⏰ 10% time]: Remove code.css

MeganSteenkamp:
Remove code.css class and move styling into tailwind config.

Test plan: Expect inline code `like this` to look unchanged.


tailwind.config.js
```diff
@@ -18,7 +10,25 @@ const tailwindConfig = {
...
       typography: {
-        DEFAULT: { css: disabledCss },
+        DEFAULT: {
+          css: {
+            'code::before': false,
+            'code::after': false,
+            'pre': false,
+            'code': false,
+            'pre code': false,
+            '*:not(pre) > code': {
```
MeganSteenkamp, line 20: @tom-blake I think when everything is markdoc we can just add styling to a component.

tom-blake (approved):

PR #1104, opened 2024-11-24
Update Farmlands connection page to use Astro components

tom-blake:
Refactoring commit only - no functional changes.

Update Farmlands connection page to use Astro components to unblock moving the MDX content to markdoc.

Test plan:
- Confirm no visual regressions in the Farmlands connection page and site Footer component


MeganSteenkamp:
@tom-blake change: I think all of the icons should have `aria-hidden=true`, this could just be added in the icon astro file. 
I'm not sure if they should all have:
```
---
const { class: className } = Astro.props;
---
```

It would make it easier if we need to use again and change size.

MeganSteenkamp (approved):

PR #1106, opened 2024-11-27
🧹 Lint tailwind classes

MeganSteenkamp:
Changes:
1. Add eslint tailwind plugin
2. Lint tailwind classes so they are in the recommended order
3. Remove invalid classnames

Test plan: Expect no visual regressions, check cards and navigation.


src/components/Error.astro
```diff
@@ -1,15 +1,16 @@
 ---
+/* eslint-disable tailwindcss/no-custom-classname */
```
MeganSteenkamp, line 2: This is because the `expanded` class is used to open and close those attributes. Have left it for now but might cull it at a later date 👀

src/components/Property.astro
```diff
@@ -1,5 +1,5 @@
-
 ---
+/* eslint-disable tailwindcss/no-custom-classname */
```
MeganSteenkamp, line 2: This is because the `expanded` class is used to open and close those attributes. Have left it for now but might cull it at a later date 👀

src/components/Prose.astro
```diff
@@ -1,13 +1,16 @@
+---
+/* eslint-disable tailwindcss/no-custom-classname */
```
MeganSteenkamp, line 2: The prose-headings class wasn't recognized even though it's valid. This will be because it's through the tailwind-typography plugin and not a normal Tailwind class.

src/components/FeaturedGuideCard.astro
```diff
@@ -4,24 +4,15 @@ const { img, title, description, slug } = Astro.props;
...
+    <div class="space-y-4 p-4">
       <p class="type-overline text-brand-accent">guide</p>
-      <h3 class="type-headline-4 hover-child link-text">{title}</h3>
+      <h3 class="type-headline-4 group-hover:underline">{title}</h3>
       <p class="type-body-2">{description}</p>
     </div>
   </a>
 </Card>
-
-<style>
```
tom-blake, line 20: 🎉

package.json
```diff
@@ -31,6 +31,7 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-standard": "^4.0.0",
+    "eslint-plugin-tailwindcss": "^3.17.5",
```
tom-blake, line 34: @MeganSteenkamp Worth considering this dependency instead - it's maintained by Tailwind

https://tailwindcss.com/docs/editor-setup#automatic-class-sorting-with-prettier

tom-blake (approved):

package.json
MeganSteenkamp: @tom-blake I did initially look at that. It would mean we would need to configure prettier and then have some duplicate rules like using single quotes or you get conflicting. I feel like having all eslint is a big plus ➕

PR #1107, opened 2024-11-27
♻️ Put used colour tokens into Tailwind config

MeganSteenkamp:
Changes:
- Remove colors.js file and define colours in Tailwind config
- Remove unused colours

Test plan: Expect no visual regressions across site including navigation and badges.


tom-blake (approved):

tailwind.config.js
```diff
@@ -31,8 +29,48 @@ const tailwindConfig = {
         },
       },
       colors: {
-        ...colors.light,
         current: 'currentColor',
+        brand: {
```
tom-blake, line 33: @MeganSteenkamp Great change! Worth considering tailwind's advice going forward - they recommend against using abstract names unless you have a really good reason to

https://tailwindcss.com/docs/customizing-colors#naming-your-colors

MeganSteenkamp: Thanks @tom-blake. I think next step I'll look at the hex codes and see if they match existing Tailwind classes, if they don't I can refactor to a literal name.

PR #1110, opened 2024-11-29
Migrate guides to markdoc

tom-blake:
Migrate guides to markdoc that didn't need additional component changes.

Test plan:
- Confirm no visual regressions for each guide


src/content/guides/example-oidc-consumer.mdoc
```diff
@@ -0,0 +1,130 @@
+---
```
tom-blake, line 1: Only changes were removing the code panel wrappers and fixing indentation

MeganSteenkamp (approved):

PR #1114, opened 2024-12-01
Use datadog for monitoring

tom-blake:
Use Datadog for monitoring to replace Hotjar. Use Vite's mode to switch environment configuration. Add environment variable to toggle Datadog integration for developers.

Test plan:
- Confirm local logs are not shown in datadog by default.
- Confirm dev and prod logs are sent to datadog
- Confirm env and version tags on logs are correctly set.

### DIFF
--- a/src/layouts/BaseLayout.astro
+++ b/src/layouts/BaseLayout.astro
@@ -40,6 +40,10 @@ const content = await Promise.all(collections.map(async page => {
 }));
 
 const navigation = Navigation.create({ nav: sideNavConfig, content });
+
+const datadogEnabled = import.meta.env.DATADOG_ENABLED;
+const env = import.meta.env.ENV;
+const version = import.meta.env.GITHUB_RUN_ID;
 ---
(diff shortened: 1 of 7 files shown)


src/layouts/BaseLayout.astro
```diff
@@ -49,12 +53,12 @@ const navigation = Navigation.create({ nav: sideNavConfig, content });
     <meta name="viewport" content="width=device-width">
     <title>{metaTitle}</title>
     <meta name="description" content={description}>
-    <meta property="img" content={`${Astro.url.origin}${img}`}>
+    <meta property="img" content={img}>
```
MeganSteenkamp, line 56: Question: Was this change intended?

MeganSteenkamp (approved):

src/layouts/BaseLayout.astro
tom-blake: @MeganSteenkamp Yep - converting these to relative links avoids complexity when switching between environments

PR #1117, opened 2024-12-02
Use a single workflow to deploy centrapay-docs

tom-blake:
Use a workflow to deploy centrapay docs instead of publishing from a branch. This allows us to consolidate our existing 4 workflows into one and use existing notifications tooling that we use for other repos.

https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow

Test plan:
- Confirm dev and prod sites are built and deployed successfully


MeganSteenkamp (approved):
Great work 🚀

PR #1132, opened 2024-12-05
🥩 Split guides and API Reference navigation

MeganSteenkamp:
Changes:
- Separate navigation for guides and API reference
- Add buttons on header bar to go between guides and API reference

Test plan: Expect to see buttons on header bar in desktop and mobile. Expect navigation to only contain relevant guide OR API content on desktop and mobile.

### DIFF
--- a/src/pages/connections/[...slug].astro
+++ b/src/pages/connections/[...slug].astro
@@ -3,6 +3,7 @@ import { getCollection } from '../../utils/getCollection';
 import customComponents from '../../utils/customComponents';
 import H2 from '../../components/H2.astro';
 import FarmlandsLayout from '../../layouts/FarmlandsLayout.astro';
+import navigation from '../../navigation/guideNavigation';
 
 export async function getStaticPaths() {
   const connections = await getCollection('connections');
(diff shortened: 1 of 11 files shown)


src/pages/connections/[...slug].astro
```diff
@@ -3,6 +3,8 @@ import { getCollection } from '../../utils/getCollection';
 import customComponents from '../../utils/customComponents';
 import H2 from '../../components/H2.astro';
 import FarmlandsLayout from '../../layouts/FarmlandsLayout.astro';
+import Navigation from '../../navigation/Navigation';
```
Comment (tom-blake) [inline src/pages/connections/[...slug].astro:6]: @MeganSteenkamp Do you need this import?

src/pages/index.astro
```diff
@@ -3,6 +3,9 @@ import BaseLayout from '../layouts/BaseLayout.astro';
 import FeaturedContentCard from '../components/FeaturedContentCard.astro';
 import { getCollection } from '../utils/getCollection';
 
+import Navigation from '../navigation/Navigation';
```
tom-blake, line 6: @MeganSteenkamp Do you need this import?

tom-blake (approved):

PR #1145, opened 2024-12-09
Add settlementBankTransferId to the response of list settlements API

Ray2126:


src/content/api/settlements.mdoc
```diff
@@ -55,6 +55,10 @@ Settlements can only be created if the Merchant has a [Settlement Config](/api/m
     The ID of the [Bank Account](/api/bank-accounts/) used for settlement.
   {% /property %}
 
+  {% property name="settlementBankTransferId" type="string" %}
+    The ID of the Bank Transfer used for settlement.
+  {% /property %}
```
subarashii-huch09, line 60: @Ray2126 
**Nitpick**  🧐
There is some inconsistency regarding the letter 'S' in the word "settlement": should it be capitalized or written in lowercase?
Screenshot 2024-12-10 at 11 31 50 AM

subarashii-huch09 (approved):

src/content/api/settlements.mdoc
Ray2126: Nice catch, they should be capitalized as they are referring to a model. I'll fix this up

PR #1152, opened 2024-12-22
Add idempotencyKey field to pay a payment request API

subarashii-huch09:
`idempotencyKey` is a new field that is required for the pay payment request API. This will allow integrators to safely retry failed requests without incurring duplicate charges.

## Changes made ##
Screenshot 2024-12-23 at 12 33 20 PM

### DIFF
--- a/src/content/api/idempotency.mdoc
+++ b/src/content/api/idempotency.mdoc
@@ -1,0 +1,9 @@
+---
+title: Idempotency
+description: Introduction to Idempotency
+nav:
+  path: API Reference
+  order: 9
+---
+Several key APIs support Idempotency. Idempotency allows requests to be safely retried without risk of performing the operation twice. When calling an API, provide an idempotency key. Then, if a connection error occurs, you can safely repeat the request using the same key without risk of the operation being performed twice.
+
--- a/src/content/api/idempotency.mdoc
+++ b/src/content/api/idempotency.mdoc
@@ -5,6 +5,8 @@ nav:
   path: API Reference
   order: 9
 ---
-Several key APIs support Idempotency. Idempotency allows requests to be safely retried without risk of performing the operation twice. When calling an API, provide an idempotency key. Then, if a connection error occurs, you can safely repeat the request using the same key without risk of the operation being performed twice.
+Several Centrapay APIs support Idempotency. Those that do, accept an idempotencyKey property in the request payload.
+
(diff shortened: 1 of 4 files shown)


Ray2126:
@subarashii-huch09 Were you planning to document our idempotency behaviour in a separate PR?

subarashii-huch09:
> @subarashii-huch09 Were you planning to document our idempotency behaviour in a separate PR?

@Ray2126  🙋🏻‍♀️ Were you thinking to create a dedicated "Idempotency" section - similar to how Stripe does it ?
Screenshot 2024-12-23 at 2 56 36 PM

Ray2126:
@subarashii-huch09 Yes, clients need to know what our APIs respond with in order to handle responses appropriately.

subarashii-huch09:
> @subarashii-huch09 Yes, clients need to know what our APIs respond with in order to handle responses appropriately.

@Ray2126  I will include our idempotency behaviour in this PR : )

src/content/api/idempotency.mdoc
```diff
@@ -0,0 +1,10 @@
+---
+title: Idempotency
+description: Introduction to Idempotency
+nav:
+  path: API Reference
+  order: 9
+---
+Several key APIs support Idempotency. Idempotency allows requests to be safely retried without risk of performing the operation twice. When calling an API, provide an idempotency key. Then, if a connection error occurs, you can safely repeat the request using the same key without risk of the operation being performed twice.
```
Ray2126, line 8: Nice start @laurencemcknight 💪

Some initial thoughts:
- We can probably remove the first line. "key APIs" doesn't mean anything to readers of this guide.

> When calling an API, provide an idempotency key.
- "APIs that support idempotency will have an `idempotencyKey` property."

- Clients should also repeat the request if a 500 occurs, not just a **network** error
- We've repeated the "risk of the operation being performed twice" line twice. Do you think we could reduce duplication here?
- Perhaps there could be a sequence diagram on this page to illustrate the example better? Maybe a client making 2 calls to the pay payment request route with the same idempotencyKey.
- Have you researched any examples in the industry that we could refer to?

laurencemcknight: Love the feedback. I've addressed the first four points.
As we discussed, re the last two points:
1. We won't add a sequence diagram as we felt is was overkill
2. Yes, Stripe.

Ray2126: Awesome, the guide is much clearer now!

PR #1159, opened 2025-01-27
add docs for list promotions by loyalty program api

bpal938:
This adds docs for the new list promotions by loyalty program api

image

test plan: 
- ensure the page is displayed correctly


src/content/api/_endpoints/promotions-list-by-loyaltyProgramId.js
```diff
@@ -0,0 +1,48 @@
...
+        rewards: [
+          {
+            assetType: 'centrapay.nzd.main',
+            amount: '500',
+          }
+        ],
+        eventType: 'payment',
+        target: {
+          type: 'count',
+          amount: 1
```
laurencemcknight, line 31: I notice that this is type int, whereas amount on line 25 is a string. Could you explain the reason for the change in type?

src/content/api/_endpoints/promotions-list-by-loyaltyProgramId.js
```diff
@@ -0,0 +1,48 @@
...
+        },
+        type: 'challenge',
+        conditions: [
+          {
+            field: 'amount',
+            value: '99',
```
laurencemcknight, line 37: If the description is "The amount of the payment must be greater than $1." then I believe this value should be 100.

Ray2126: 99 is correct. `greater-than` operator is not inclusive

Ray2126 (approved):

src/content/api/_endpoints/promotions-list-by-loyaltyProgramId.js
bpal938: @laurencemcknight It was just the same as the response from the similar list promotions by account Id api, I can double check the values here

laurencemcknight: Disregard, I see that it has type `number` (https://docs.centrapay.com/api/promotions/#promotion-target-model)

src/content/api/_endpoints/promotions-list-by-loyaltyProgramId.js
Ray2126: Sorry, Laurence is correct. Since the payment amount must be greater than $1 (100 cents), we should put `100`

src/content/api/_endpoints/promotions-list-by-loyaltyProgramId.js
bpal938: @laurencemcknight It seems like this is just how promotions are structured, I am guessing it is because the other value is a monetary value so we store it as a string, while this is always a whole number

PR #1162, opened 2025-01-30
create applet api

seliwe:
Added public docs for the [Create Applet API](https://www.notion.so/centrapay/Create-Applet-183a9ab17e80801592aced8e973a2c4c?pvs=4)


src/content/api/applets.mdoc
```diff
@@ -0,0 +1,89 @@
...
+
+    {% property name="accountId" type="string" required=true %}
+      The [Account](/api/accounts/) that will own the Applet.
+    {% /property %}
+
+    {% property name="mediaUploadId" type="string" %}
+      The id of the [Media Upload](/api/media-uploads/) image of the Applet.
+    {% /property %}
+
+    {% property name="test" type="boolean" %}
```
Ray2126, line 73: @seliwe Should the "test" flag be derived from the liveness of the account? We were saying that live and test resources should be separated completely: https://www.notion.so/centrapay/Separate-test-and-live-resources-completely-6cd466d9e3a145c88b6f340ba9493002

seliwe: @Ray2126 nice, have removed

Ray2126 (approved):

PR #1172, opened 2025-02-24
add docs for list applets api

bpal938:
Test Plan: Ensure the page loads correctly

image
image


src/content/api/_endpoints/applets-list.js
```diff
@@ -0,0 +1,23 @@
...
+  path: '/api/accounts/Jaim1Cu1Q55uooxSens6yk/applets',
+  request: {
+    headers: {
+      'X-Api-Key': '<TOKEN>',
+      'Content-Type': 'application/json',
+    },
+  },
+  response: {
+    items: [
+      {
```
Ray2126, line 12: Did you want to add extensions to this response too?

Ray2126 (approved):

laurencemcknight (approved):

src/content/api/_endpoints/applets-list.js
bpal938: Good point, they were also missing from the create api so I added in the model for that as well

PR #1181, opened 2025-03-17
Add error for create loyalty program membership on inactive status

Ray2126:


src/content/api/loyalty-program-memberships.mdoc
```diff
@@ -79,5 +79,9 @@ A Loyalty Program Membership represents an Account's membership to a Loyalty Pro
     {% error code="403" message="LOYALTY_PROGRAM_ACCOUNT_LIVENESS_MISMATCH" %}
      The account cannot be test if the loyalty program is not test.
     {% /error %}
+
+    {% error code="403" message="LOYALTY_PROGRAM_INACTIVE" %}
+     The loyalty program does not have status of `active`.
```
laurencemcknight, line 84: Suggestion: "Memberships cannot be created for loyalty programs with an 'inactive' status."
I think this helps clarify the underlying business logic. LMK what you think.

Ray2126: Just to clarify, loyalty programs do not have an `inactive` status. I'll flip the wording to say "Memberships can only be created for loyalty programs with an `active` status."

laurencemcknight (approved):

PR #1183, opened 2025-03-18
Add error for create applet when extensions are invalid

Ray2126:


src/content/api/applets.mdoc
```diff
@@ -114,6 +114,10 @@ An applet is a discoverable container of functions and features that users can a
     {% error code="403" message="MEDIA_UPLOAD_INVALID" %}
       The media upload id does not reference a media upload.
     {% /error %}
+
+    {% error code="403" message="APPLET_EXTENSIONS_INVALID" %}
+      One of more of the resources referenced in the `extensions` array either do not exist or are invalid.
```
laurencemcknight, line 119: Are we able to provide details on what can cause an extension to be invalid?

Ray2126: I've updated it now, what do you think?

laurencemcknight: To summarise our conversation: I'm happy for you to revert the description change so long as a detailed error reason is returned to the client alongside the APPLET_EXTENSIONS_INVALID message.

laurencemcknight (approved):

PR #1189, opened 2025-04-03
Add documentation for notifications & payment/refund notifications

Ray2126:


src/navigation/apiNavigation.js
```diff
@@ -15,6 +15,7 @@ const nav = [
   { title: 'Loyalty Programs' },
   { title: 'Media Uploads' },
   { title: 'Merchants' },
+  { title: 'Notifications' },
```
tom-blake, line 18: @Ray2126 Change: This should be called "Webhooks" to match standard terminology that developers expect

src/content/api/notifications.mdoc
```diff
@@ -0,0 +1,44 @@
+---
```
tom-blake, line 1: @Ray2126 I had expected us to write a guide for developers (and BNZ). Is this coming too?

API reference isn't enough. Developers need to know how to subscribe to webhooks, how to authenticate requests, best practices etc

tom-blake: Good examples are Stripe and Github

src/content/api/notifications.mdoc
```diff
@@ -0,0 +1,44 @@
...
+  {% property name="id" type="string" %}
+    Notification's unique identifier.
+  {% /property %}
+
+  {% property name="type" type="string" %}
+    The event that occurred that triggered the Notification. See below for the list of types.
+  {% /property %}
+
+  {% property name="data" type="string" %}
+    The associated metadata.
```
tom-blake, line 22: @Ray2126 Suggestion: Provide more context here - this is a very generic description

src/content/api/notifications.mdoc
```diff
@@ -0,0 +1,44 @@
...
+  {% property name="test" type="boolean" %}
+    Whether the Notification is for testing purposes.
+  {% /property %}
+
+  {% property name="createdAt" type="timestamp" %}
+    When the notification was created.
+  {% /property %}
+{% /properties %}
+
+## Notification Types
```
tom-blake, line 38: @Ray2126 Change: Inline event payloads for each type to reduce indirection for readers and coupling to our activity models

Example: [Github](https://docs.github.com/en/webhooks/about-webhooks) - this is actually part of their guide, not their API reference

src/navigation/apiNavigation.js
Ray2126: They are called Notifications in our system though. If we expose APIs to retrieve notifications, I would expect them to be called "List Notifications", "Get Notification" etc. rather than "List Webhooks".

They are called [Events](https://docs.stripe.com/api/events) in Stripe so I'm not sure we need to conform to the Webhook terminology.

src/content/api/notifications.mdoc
Ray2126: Yep sure, I had thought BNZ were only after the fields on the Notification for now. I can add a guide as well

src/content/api/notifications.mdoc
Ray2126: I disagree, I think they should be coupled to our other models. This way, we can provide a consistent schema for the models that we expose to integrators.

Example: [Stripe](https://docs.stripe.com/api/events/types) - they link out to their other models for each Event type

src/navigation/apiNavigation.js
Ray2126: But the API to subscribe might be called "Create Webhook"? Or "Create Notification Subscription"?

I like keeping the language here consistent with our system to avoid confusion. Thoughts?

src/content/api/notifications.mdoc
Ray2126: Added a guide now :)

src/navigation/apiNavigation.js
tom-blake: @Ray2126 Events are the payload, not the feature. The webhook terminology is important - I don't know of a single tool that doesn't use it and we actively look for it ourselves when evaluating new vendors 👍  

Stripe's webhook guide is here -> https://docs.stripe.com/webhooks

tom-blake: The language I had in mind was:
- Developers create a **webhook endpoint** in their system
- A webhook endpoint can be registered with Centrapay by **subscribing**  it to one or more **events**
- Once subscribed, a webhook endpoint will re./ive **notifications** (requests) when those events occur
- A notification contains an the event payload and other metadata (delivery status etc)

So "Listing notifications" would list all deliveries to your webhook endpoint

Ray2126: As discussed, we will separate Events from Notifications in API reference section.

src/content/api/notifications.mdoc
Ray2126: As discussed, we agree that notifications should reference the model to provide consistency

src/content/api/notifications.mdoc
```diff
@@ -0,0 +1,45 @@
...
+  {% property name="id" type="string" %}
+    Notification's unique identifier.
+  {% /property %}
+
+  {% property name="type" type="string" %}
+    The [notification type](/api/notifications#notification-types).
+  {% /property %}
+
+  {% property name="data" type="string" %}
+    The associated metadata. Typically, this would be the resource state at the time of change.
```
tom-blake, line 22: Suggestion: This is quite technical language. We could say something along the lines of "the details of the event" or similar to make it easier to read

tom-blake (approved):

PR #1197, opened 2025-04-14
add repeatable flag to create promotion docs

bpal938:
This adds docs for the new repeatable flag available when creating promotions

Test Plan: Ensure the page displays as intended

image

### DIFF
--- a/src/content/api/promotions.mdoc
+++ b/src/content/api/promotions.mdoc
@@ -201,7 +201,11 @@ Promotions are a mechanism to reward accounts for completing certain actions.
     {% /property %}
 
     {% property name="supplyLimit" type="string" %}
-    The maximum number of [Promotion Memberships](/api/promotion-memberships) that can be created.
+      The maximum number of [Promotion Memberships](/api/promotion-memberships) that can be created.
+    {% /property %}
+
+    {% property name="repeatable" type="boolean" %}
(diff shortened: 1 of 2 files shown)


src/content/api/promotions.mdoc
```diff
@@ -201,7 +201,11 @@ Promotions are a mechanism to reward accounts for completing certain actions.
     {% /property %}
 
     {% property name="supplyLimit" type="string" %}
-    The maximum number of [Promotion Memberships](/api/promotion-memberships) that can be created.
+      The maximum number of [Promotion Memberships](/api/promotion-memberships) that can be created.
+    {% /property %}
+
+    {% property name="repeatable" type="boolean" %}
+      Flag indicating whether the promotion will reward itself upon completion, making it repeatable.
```
laurencemcknight, line 208: "the promotion will reward itself upon completion" IMO this is an implementation detail and as such should be left out of the public docs. However, LMK if you think its inclusion is helpful for integrators.

Ray2126: I don't think it is an implementation detail because I believe we will be returning the `promotionId` in the `rewards` array.

Ray2126 (approved):

src/content/api/promotions.mdoc
Ray2126: Suggestion: You could call out that providing the repeatable flag means that members can complete the promotion an unlimited number of times.

PR #1200, opened 2025-04-17
Add accountId to the merchant search query params

Ray2126:
Screenshot 2025-04-17 at 12 13 43 PM


src/content/api/_endpoints/merchants-search.js
```diff
@@ -2,9 +2,6 @@ export default {
   method: 'GET',
   path: '/api/merchants/search',
   request: {
-    headers: {
```
Ray2126, line 5: Also removed this because the endpoint does not require auth.

laurencemcknight (approved):

PR #1209, opened 2025-05-28
Add category code to the merchant search endpoint

Ray2126:


src/content/api/merchants.mdoc
```diff
@@ -127,6 +127,14 @@ which define the payment methods available for a Payment Request.
   {% property name="location" type="location" %}
     Physical location of the merchant
   {% /property %}
+
+  {% property name="accountId" type="string" %}
```
laurencemcknight, line 131: Suggestion: In future it's worth calling out why changes like this are included

laurencemcknight (approved):

bpal938 (approved):

src/content/api/merchants.mdoc
Ray2126: Yep sure, my bad. For this one, I just noticed it was missing so added it in - but yes you're right. It's not related to the PR specifically.

PR #1220, opened 2025-06-17
update docs for create and get wallet

bpal938:
Test Plan: 
- Ensure the docs show as `/wallet-assets` for the route to call

image


Ray2126 (approved):

Ray2126:
Should we also change the page title? Can also be done in a follow-up PR when we actually document the new wallets.

PR #1223, opened 2025-06-25
add docs for get applet api

bpal938:
Adds docs for the new get applet API

Test Plan: 
- Ensure the docs show as expected

image

### DIFF
--- a/src/content/api/applets.mdoc
+++ b/src/content/api/applets.mdoc
@@ -134,3 +134,15 @@ An applet is a discoverable container of functions and features that users can a
 {% /endpoint %}
 
 ---
+
+{% endpoint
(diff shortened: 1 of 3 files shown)


src/content/api/applets.mdoc
```diff
@@ -134,3 +134,15 @@ An applet is a discoverable container of functions and features that users can a
...
 
 ---
+
+{% endpoint
+  path="/api/applets/{id}"
+  filename="applets-get"
+%}
+  ## Get Applet {% badge type="experimental" /%}
+
+  Returns an Applet based on its id
```
Ray2126, line 144: "Get an Applet based on it's ID"?

PR #1224, opened 2025-06-26
Add line items to the pay payment request route

Ray2126:


bpal938 (approved):

src/content/api/payment-requests.mdoc
```diff
@@ -679,6 +679,9 @@ Payment Activities are created when a Payment Request has been `created`, `paid`
     {% property name="externalPaymentRef" type="string" %}
       An external reference to the payment. Required when `assetType` is `farmlands.nzd.*`.
     {% /property %}
+    {% property name="lineItems" type="array" %}
+      The [Line Items](#line-item-model) associated with this payment attempt. This field is only allowed for `centrapay.token.*` payments. If omitted for `centrapay.token.*` payments, the highest-value valid line item from the payment request will be used by default.
```
laurencemcknight, line 683: I think I'd be worthwhile mentioning the restrictions that:
lineItems.length == 1
and 
lineItem[0].qty == 1

Ray2126: This is only true for tokens of `type: product`. If it were true for all tokens, we woul have made lineItems a single item instead of an array. I'm keen to leave it as it is to keep the door open for other token types that may redeem more than 1 token at a time.

PR #1231, opened 2025-07-15
Docs for list wallets

fedch:
Docs for listing a merchant’s supported wallets
Screenshot 2025-07-15 at 12 29 58

### DIFF
--- a/src/content/api/_endpoints/wallets-list.js
+++ b/src/content/api/_endpoints/wallets-list.js
@@ -1,0 +1,24 @@
+export default {
+  method: 'GET',
+  path: '/api/merchants/{merchantId}/wallets',
+  request: {
+    headers: {
+      'X-Api-Key': '<TOKEN>',
+      'Content-Type': 'application/json',
+    },
+  },
+  response: [
+    {
+      id: '1234',
+      name: 'payap',
+      accountId: 'abc123',
+      connectUrl: 'https://app.payap.com/join',
+      payUrl: 'https://app.payap.com/pay',
+      assetTypes: [
+        { id: 'quartz.nzd' },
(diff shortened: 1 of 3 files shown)


src/content/api/_endpoints/wallets-list.js
```diff
@@ -0,0 +1,24 @@
+export default {
+  method: 'GET',
+  path: '/api/merchants/{merchantId}/wallets',
```
bpal938, line 3: Instead of {merchantId} can you put an id string (doesn't need to be a real id)

src/content/api/_endpoints/wallets-list.js
```diff
@@ -0,0 +1,24 @@
...
+  response: [
+    {
+      id: '1234',
+      name: 'payap',
+      accountId: 'abc123',
+      connectUrl: 'https://app.payap.com/join',
+      payUrl: 'https://app.payap.com/pay',
+      assetTypes: [
+        { id: 'quartz.nzd' },
+        { id: 'airpoints' },
```
bpal938, line 19: Could we replace this with a assetType that is currently real

src/content/api/wallets.mdoc
```diff
@@ -104,3 +104,17 @@ This endpoint allows you to create a [Wallet](#wallet-model).
 {% /endpoint %}
 
 ---
+
+## List Wallets {% badge type="experimental" /%}
+This endpoint allows you to retrieve a list of [Wallets](#wallet-model) that a merchant supports.
```
bpal938, line 109: Suggestion: add something about how we are fetching the wallets based on the merchants supported assets

Ray2126 (approved):

PR #1252, opened 2025-08-27
Accordion component on devices page

alexBinns:
This adds a accordion component to hide the devices on the device page.

image

### DIFF
--- a/src/components/Accordion.astro
+++ b/src/components/Accordion.astro
@@ -1,0 +1,41 @@
+---
+import Carat from "./icons/Carat.vue";
+const { title } = Astro.props;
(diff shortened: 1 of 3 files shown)


src/components/Accordion.astro
```diff
@@ -0,0 +1,41 @@
+---
```
tom-blake, line 1: Why do we need this?

I'd like to aim for vanilla markdown and avoid complex components. Makes it easier to put a CMS on top later for anyone to edit

alexBinns: Open to suggestions

alexBinns: Where was that branch you were working on when you were doing your vibe coding to make each route its own page?

PR #1254, opened 2025-08-27
Add ontempo to to compatible solutions

alexBinns:
Adding ontempo to compatible solutions


src/content/guides/compatible-solutions.mdoc
```diff
@@ -17,6 +17,8 @@ Centrapay is currently integrated with a wide range of payment and point-of-sale
 - G6-400
 - G6-500
 
+### Ontempo
```
IsaacMurtagh, line 20: Are they a POS or online solution? Might be worth adding some detail in the bullet point

IsaacMurtagh (approved):

PR #1256, opened 2025-09-01
Remove 'test' from asset program

cwenborn:
Removes 'test' param from endpoint as this is instead filtered by account liveness.


src/content/api/asset-programs.mdoc
```diff
@@ -29,10 +29,6 @@ An Asset Program is provided by an asset issuer as a mechanism that enables busi
     The [Account](/api/accounts/) that owns the Asset Program.
   {% /property %}
 
-  {% property name="test" type="boolean" %}
```
IsaacMurtagh, line 32: This is still required, we're not removing the test property, it just shows up depending on your own accounts liveness.

cwenborn: I've re-added this property 😊

PR #1331, opened 2026-04-21
Connect flow guide

alexBinns:
This is a revised connect flow guide that makes the page into 1 with the mermaid diagrams for wallet and pos in different tabs


src/content/guides/connect-flow.mdoc
```diff
@@ -0,0 +1,189 @@
...
+
+- Validating the patron's identity before applying discounts or offers
+- Retrieving and applying promotional offers and discounts
+- Redeeming loyalty tokens as partial payment
+
+The POS [creates a Connection](/api/connections#create-connection) and displays the Connection URL as a QR code. When the patron scans it, they authorises the Connection through the wallet, linking the wallet to the transaction. The wallet then determines eligibility for discounts and offers before any payment is taken, and the remaining balance is collected by the POS.
+
+## Payment Flow
+
+{% tabs %}
```
Ray2126, line 23: Suggestion: These diagrams are a bit too confusing and complex IMO. Could we cut back on some of the edge cases to simplify it?

src/content/guides/connect-flow.mdoc
```diff
@@ -0,0 +1,189 @@
...
+
+**Wallet:** Implement the authorise connection endpoint that wallet calls when the patron scans the QR code. Use this call to validate the patron's identity and link them to the connection. Once validated, call Centrapay to [authorise the connection](/api/connections#authorize-connection) and return a `200 OK` to the wallet.
+
+## Requesting payment
+
+Once the connection is authorised, the POS creates a Payment Request tied to the patron's session.
+
+**POS:** [Create a Payment Request](/api/payment-requests#create-a-payment-request) with the `connectionId` from the Connection. This allows Centrapay to associate the transaction with the patron's wallet so offers can be applied.
+
+**Wallet:** Implement an endpoint that the wallet can call to retrieve the Payment Request by `connectionId`. Use this to fetch the payment request details from Centrapay and return them to the app, including line items that will be used to calculate available offers.
```
Ray2126, line 143: Suggestion: I feel we don't need to tell the wallet integrator to "implement an endpoint". We could make this description simpler by just saying 

```
Call the Get Payment Request by connectionId API to fetch the payment request details from Centrapay. The payment request will contain line items that will be used to calculate available offers.
```

src/content/guides/connect-flow.mdoc
```diff
@@ -0,0 +1,189 @@
...
+
+## Applying Offers and Processing Payment
+
+The POS and wallet work in parallel — the POS polls for the payment request status while the wallet calculates and applies offers.
+
+**POS:** Begin polling the Payment Request for status changes. If Centrapay returns an updated Payment Request with discounts applied, update the cart accordingly. Once offers have been applied, display the Payment Request QR code and request the remaining balance from the patron via cash, EFTPOS, or a Centrapay payment. If no discounts were applied and the patron pays by a non-Centrapay method, void the Payment Request after collecting payment.
```
Ray2126, line 149: > If no discounts were applied and the patron pays by a non-Centrapay method, void the Payment Request after collecting payment.

This is false.

If the user connected the wallet, did not apply discounts and paid with cash, we will still let the wallet integrator know that the payment happened so they can increment any promotion progress. This is the behaviour with Coke and Fizz Club.

Instead, we could say:
"If the connection was not authorised, void the Payment Request after collecting payment".

src/content/guides/connect-flow.mdoc
```diff
@@ -0,0 +1,189 @@
...
+
+**Wallet:** Implement the endpoints Centrapay calls to retrieve and apply offers:
+
+- **Get offers** — calculate available offers based on the payment request's line items and return them to the wallet.
```
Ray2126, line 153: Get Offer is not an endpoint that Centrapay calls.

All we care about is the wallet integrator making payments on the payment request. The only endpoint they need to build for payments is the Uplink API Pay endpoint.

Ray2126 (approved):
Approving to unblock. Please address the comments and I did notice a few typos/sentences that could be worded better. Might be worth doing a proofread with AI.

PR #1338, opened 2026-05-06
Fix mobile hamburger menu

alexBinns:
When making the changes to add the merchant services section the hamburger menu on mobile was pushed off the edge. This PR puts it back. It also moves the section selection for mobile into the hamburger menu as there isn't much real estate width wise on mobile.

image
image

Testing Steps:
* Go to page on mobile see that the hamburger menu is there and click and see the merchant selection


Ray2126 (approved):
Make sure to check that nothing has regressed on the desktop view too!

PR #1339, opened 2026-05-08
Use open api spec

alexBinns:
This PR updates the create payment request route to use open api spec. 
This is technically not fully open api spec as using variables in the example is not supported by the spec.
It also adds Spectral as part of the lint check
Testing Steps:
* See that there are no changes to the create payment request section on the payment request page


Linlin-Liang (approved):

src/content/api/payment-requests.mdoc
```diff
@@ -472,117 +472,11 @@ Payment Activities are created when a Payment Request has been `created`, `paid`
...
-    {% error code="403" message="CONNECTION_MERCHANT_MISMATCH" %}
-      The connection is not for the same merchant as the payment request.
-    {% /error %}
-    {% error code="403" message="CONNECTION_CREATED_BY_MISMATCH" %}
-      The connection was not created by the same caller as the payment request.
-    {% /error %}
-  {% /properties %}
-
-{% /endpoint %}
+{% /openApiEndpoint %}
```
tom-blake, line 479: @alexBinns We should be able to put this metadata in the openapi spec and get rid of the .mdoc files completely

src/content/api/openapi/index.yaml
```diff
@@ -0,0 +1,31 @@
...
+          type: string
+          description: The error code.
+    LineItem:
+      $ref: './models/line-item.yaml'
+    PaymentRequest:
+      $ref: './models/payment-request.yaml'
+    MerchantConfig:
+      $ref: './models/merchant-config.yaml'
+    Merchant:
+      $ref: './models/merchant.yaml'
```
tom-blake: @alexBinns Can we move these model references inline inside the `create-payment-request.yaml` file? Keeps endpoint details better encapsulated and stops this file becoming HUGE over time

src/content/api/openapi/endpoints/create-payment-request.yaml
```diff
@@ -0,0 +1,125 @@
+x-method: POST
+x-path: /api/payment-requests
+operationId: createPaymentRequest
+summary: Create a Payment Request
+tags:
+  - Payment Requests
```
tom-blake: Suggest using kebab case tags so they're machine readable `payment-requests`

src/content/api/openapi/endpoints/create-payment-request.yaml
```diff
@@ -0,0 +1,125 @@
...
+          basketAmount:
+            $ref: '../models/payment-request.yaml#/properties/basketAmount'
+          connectionId:
+            $ref: '../models/payment-request.yaml#/properties/connectionId'
+          paymentLinkId:
+            $ref: '../models/payment-request.yaml#/properties/paymentLinkId'
+      examples:
+        default:
+          $ref: '../models/payment-request.yaml#/x-examples/PaymentRequestBody'
+responses:
```
tom-blake: @alexBinns Keen to include all possible status codes in each route - 401, 429, 500 etc

https://docs.centrapay.com/api/http-status-codes/

Maybe there's a way to define these centrally so that all routes automatically inherit them?

src/components/OpenApiEndpoint.astro
```diff
@@ -0,0 +1,153 @@
...
+
+const { operationId } = Astro.props;
+
+const schemasCollection = await getCollection('openapiSchemas');
+
+const schemasMap = Object.fromEntries(
+  schemasCollection.map(entry => [entry.id.split('/').pop(), entry.data])
+);
+
+function resolveAll(obj) {
```
tom-blake: @alexBinns Change: These code is super gnarly to read (clanker?). Would benefit from some proper OOP classes so that it's easier to maintain and extend

src/content/api/openapi/index.yaml
```diff
@@ -0,0 +1,31 @@
+openapi: 3.0.0
```
tom-blake, line 1: @alexBinns Change - we need a script that can lint our OpenAPI spec (both locally and in our CI pipeline) so that it remains valid after this PR is merged and so we can give it to integrators and pen testers. Lots of libraries out there for this

src/content/api/payment-requests.mdoc
alexBinns: Which metadata? All the information relevant to the endpoint is currently in the open api spec files. I was thinking we can move to separate files once we have migrated all the endpoints. The main goal of this PR is to get one endpoint migrated and the pattern established.

src/content/api/openapi/index.yaml
Linlin-Liang: I have added Spectral

src/content/api/openapi/endpoints/create-payment-request.yaml
Linlin-Liang: the tag is updated

src/content/api/openapi/index.yaml
Linlin-Liang: 4 model references are all moved

src/components/OpenApiEndpoint.astro
Linlin-Liang: The function is split into multiple small functions.

src/content/api/openapi/endpoints/create-payment-request.yaml
Linlin-Liang: response.yaml file which includes the common status code is added. getErrors() will return both response in each route and common status code.

package.json
```diff
@@ -20,6 +21,7 @@
     "@hbsnow/rehype-sectionize": "^1.0.7",
     "@headlessui/vue": "^1.6.0",
     "@intercom/messenger-js-sdk": "^0.0.18",
+    "@stoplight/spectral-cli": "6.16.1",
```
tom-blake, line 23: @Linlin-Liang Sorry to be a pain - can you write an ADR for this?

https://www.notion.so/centrapay/Write-an-ADR-for-new-dependencies-b34d6e63f50c4e1a9192e09a197aac5f

Keen to understand what options there are

src/content/api/openapi/endpoints/create-payment-request.yaml
```diff
@@ -0,0 +1,118 @@
...
+            basketAmount:
+              $ref: '../models/payment-request.yaml#/properties/basketAmount'
+            connectionId:
+              $ref: '../models/payment-request.yaml#/properties/connectionId'
+            paymentLinkId:
+              $ref: '../models/payment-request.yaml#/properties/paymentLinkId'
+        examples:
+          default:
+            $ref: '../models/payment-request.yaml#/x-examples/PaymentRequestBody'
+  responses:
```
tom-blake, line 68: @Linlin-Liang Looks like we're missing `200` response documentation. Might belong in @responses.yaml?

package.json
```diff
@@ -3,7 +3,8 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "lint": "eslint .",
+    "lint": "eslint . && yarn lint:openapi",
+    "lint:openapi": "spectral lint src/content/api/openapi/index.yaml",
```
tom-blake: Suggestion - just make this part of yarn lint directly 👍 

```suggestion
```

.spectral.yaml
```diff
@@ -0,0 +1,7 @@
+extends:
```
tom-blake, line 1: It looks like the CI pipeline isn't failing despite warnings from the lint. Can we make this super strict so the pipeline fails?

src/content/api/openapi/endpoints/create-payment-request.yaml
Linlin-Liang: 200 is added back to route yaml file since each route will have a different example( in this case "PaymentRequestResponse")

package.json
Linlin-Liang: ADR: https://app.notion.com/p/centrapay/Add-Spectral-to-centrapay-docs-3ada9ab17e808083a6dafb215113df17?v=a08c6f5ea24b4a93beed878a46a280ec&source=copy_link

PR #1354, opened 2026-06-24
add metadata endpoint and property

Linlin-Liang:
This is to add documentation for the new endpoint to set metadata on patron codes. 

Testing step:
Open centrapay doc
Go to patron code page
Check that set metadata endpoint is added, and the metadata property is added for other create and get endpoints as well.


src/content/api/patron-codes.mdoc
```diff
@@ -149,3 +153,26 @@ be used with Payment Requests that will have a liveness of 'test'. The [Asset Ty
...
+{% endpoint
+  path="/api/patron-codes/{patronCodeId}/metadata"
+  filename="patron-codes-set-metadata"
+%}
+  ## Set Patron Code Metadata
+
+  This endpoint allows you to set metadata on a Patron Code.
+
+  {% properties %}
+    {% property name="metadata" type="object" required=true %}
```
alexBinns, line 170: I feel like we should explain a bit more thoroughly how to add, remove and update

Linlin-Liang: I added more description on how to set metadata. Let me know if needs further changes.

alexBinns (approved):

PR #1355, opened 2026-06-24
Update e-commerce guide structure

hanface:
### Context
Update the structure of the e-commerce guide to bring all the prerequisites to the top, and bring more examples inline to make it easier for both developers and agents to scan the guide and build an integration.

### Test plan
- ensure doc still contains all revelant information


src/content/guides/ecommerce-website.mdoc
```diff
@@ -8,28 +8,79 @@ nav:
...
+    });
+    const json = await response.json();
+    res.status(response.status).json(json);
+  } catch (error) {
+    console.error('Failed to create payment request:', error);
+    res.status(500).json({ error: 'Failed to create payment request.' });
+  }
+});
+
+app.get('/api/payment-requests/:paymentRequestId', async (req, res) => {
```
alexBinns, line 69: If we are showing the get we should add the polling logic maybe?

hanface: Does it need polling? in the code example it doesn't have any. I have not tested the redirect, but the popup doesn't need it as the centrapay page handles it

alexBinns: Yea the popup shouldn't but the redirect should? Can they detect if the customer comes back from a redirect?

alexBinns: What if the customer just clicks back?

alexBinns: 🤷

src/content/guides/ecommerce-website.mdoc
```diff
@@ -87,9 +138,28 @@ autonumber
 
     Centrapay will return a `url` which you are expected to redirect the customer to.
 
+    ```javascript title="Client-side — Browser"
```
alexBinns, line 141: Are these titles for the ai? They are not rendering on the page.

alexBinns (approved):

src/content/guides/ecommerce-website.mdoc
hanface: Good thinking - I can add this in the future. For now I am including it in the agent skills.md  to be able to handle edge cases :)

src/content/guides/ecommerce-website.mdoc
hanface: Yeah, it's just using a common naming pattern so it's easier to distinguish FE & BE

PR #1368, opened 2026-07-28
Add merchantId and merchantAccountId on the promotion conditions

esennik:
For [this](https://github.com/centrapay/kari/pull/1588) and [this](https://github.com/centrapay/kari/pull/1590) createPromotion changes

### DIFF
--- a/src/content/api/promotions.mdoc
+++ b/src/content/api/promotions.mdoc
@@ -151,7 +151,7 @@ Promotions are a mechanism to reward accounts for completing certain actions.
   {% /property %}
 
   {% property name="type" type="string" %}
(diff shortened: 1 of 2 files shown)


src/content/api/promotions.mdoc
```diff
@@ -151,7 +151,7 @@ Promotions are a mechanism to reward accounts for completing certain actions.
   {% /property %}
 
   {% property name="type" type="string" %}
-    The type of condition. Can be `participation-limit`.
+    The type of condition. Can be `participation-limit`, `merchantId`, or `merchantAccountId`. Use `merchantId` or `merchantAccountId` to match on that field without also specifying `field`.
```
Ray2126: Let's remove:

> Use `merchantId` or `merchantAccountId` to match on that field without also specifying `field`.

I don't think the sentence makes sense. It's probably not needed.

Ray2126 (approved):

